PR #31, opened 2018-08-24
Update price range filter facet to use range component

lucasecdb:
Price ranges should be using a separate query parameter, instead of the `rest` used for brands and additional categories. Also, the price range facet should be using a range slider component, instead of a list of the facets options.

[Workspace](https://lucas--storecomponents.myvtex.com/eletronicos/smartphones).

#### Issues

* [ ] Update README.md with entrypoint component props
* [x] Update price range filter to use a separate query parameter (priceRange)
* [x] Create slider component to replace price range

### DIFF
--- a/react/components/FilterOptionTemplate.js
+++ b/react/components/FilterOptionTemplate.js
@@ -48,7 +48,7 @@ export default class FilterOptionTemplate extends Component {
     const { selected, title, collapsable, children, filters } = this.props
     const { open } = this.state
 
-    // keep old behaviour
+    // Backward-compatible support
     if (typeof children === 'function' && !filters.length) {
       return null
     }
--- a/react/components/FilterOptionTemplate.js
+++ b/react/components/FilterOptionTemplate.js
@@ -10,10 +10,13 @@ import Arrow from '../images/Arrow'
  */
 export default class FilterOptionTemplate extends Component {
   static propTypes = {
-    /** Filters to be shown */
-    filters: PropTypes.arrayOf(PropTypes.object).isRequired,
-    /** Function to handle filter rendering */
-    children: PropTypes.func.isRequired,
+    /** Filters to be shown, if no filter is provided, treat the children as simple node */
+    filters: PropTypes.arrayOf(PropTypes.object),
+    /** Function to handle filter rendering or node if no filter is provided */
+    children: PropTypes.oneOfType([
+      PropTypes.func,
+      PropTypes.node,
+    ]).isRequired,
     /** Title */
     title: PropTypes.string.isRequired,
     /** Whether collapsing is enabled */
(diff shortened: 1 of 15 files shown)


react/components/FilterOptionTemplate.js
```diff
@@ -31,11 +34,22 @@ export default class FilterOptionTemplate extends Component {
...
+    return filters.map(children)
+  }
+
   render() {
-    const { selected, children, filters, title, collapsable } = this.props
+    const { selected, title, collapsable, children, filters } = this.props
     const { open } = this.state
 
-    if (!filters.length) {
+    // keep old behaviour
```
andreldsa: Change to "Backward-compatible support"

react/components/Range.js
```diff
@@ -0,0 +1,297 @@
...
+    step: PropTypes.number,
+    disabled: PropTypes.bool,
+    initialValues: PropTypes.shape({
+      left: PropTypes.number,
+      right: PropTypes.number,
+    }),
+  }
+
+  static defaultProps = {
+    min: 0,
```
vitoriaHeliane: Não seria melhor usar constates?

react/components/Range.js
```diff
@@ -0,0 +1,297 @@
...
+            }}
+            onDragStart={this.handleDragStart}
+            position="left"
+            active={this.state.dragging === 'left'}
+            value={this.state.values.left}
+          />
+          <RangeCircle
+            style={{
+              transform: `translateX(-${right}px) translateX(100%)`,
+              right: 0,
```
vitoriaHeliane: Por que não usa a classe ```right-0``` do tachyons?

react/components/Range.js
```diff
@@ -0,0 +1,297 @@
...
+                height: 3,
+                left,
+                right,
+              }}
+            />
+          </div>
+          <RangeCircle
+            style={{
+              transform: `translateX(${left}px) translateX(-100%)`,
+              left: 0,
```
vitoriaHeliane: Por que não usa a classe ```left-0``` do tachyons?

react/components/Range.js
lucasecdb: why?

react/components/Range.js
```diff
@@ -0,0 +1,297 @@
...
+    this.updateLayout()
+  }
+
+  render() {
+    const { disabled } = this.props
+    const { left, right } = this.state.translate
+
+    return (
+      <div className="ph4">
+        <div className="vtex-range w-100 relative pv4" style={{ height: 48 }}>
```
ClaudivanFilho: you can use h2 `height: 32px` of tachyons and set a padding higher, like pv7

react/components/RangeCircle.js
```diff
@@ -0,0 +1,62 @@
...
+
+RangeCircle.defaultProps = {
+  active: false,
+  disabled: false,
+  value: 0,
+  className: '',
+}
+
+RangeCircle.propTypes = {
+  className: PropTypes.string,
```
ClaudivanFilho: Doc :eyes:

react/components/Range.js
lucasecdb: It doesn't work, I need at least 48px of height for the range to not visually break.

lucasecdb:
[Update price range filter facet to use range component](https://trello.com/c/35JXjpZf/26-update-price-range-filter-facet-to-use-range-component)

PR #41, opened 2018-09-13
Fix incorrect specification filter being removed

lucasecdb:
Fixes condition in the break loop when removing the filter.

Fixes inccorect specification filter being removed when there were more than one instance of it in the url.

[Workspace](https://lucas--storecomponents.myvtex.com/eletronicos/smartphones), select an specification filter then another, then remove the first one you selected.

#### Types of changes

### DIFF
--- a/react/constants/SearchHelpers.js
+++ b/react/constants/SearchHelpers.js
@@ -101,7 +101,7 @@ function removeFilter(map, rest, { type, slug, pagesPath }) {
 
     if (symbol === categoryMapSymbol && skip > 0) {
       skip--
-    } else if (count === restIndex) {
+    } else if (count >= restIndex) {
       break
     } else {
       count++
--- a/react/constants/SearchHelpers.js
+++ b/react/constants/SearchHelpers.js
@@ -73,14 +73,14 @@ function restMapped(rest, map) {
 }
 
 function removeFilter(map, rest, { type, slug, pagesPath }) {
-  let skip = 0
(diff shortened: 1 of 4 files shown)


react/constants/SearchHelpers.js
```diff
@@ -93,28 +93,42 @@ function removeFilter(map, rest, { type, slug, pagesPath }) {
...
+
+    const mapWithoutCategories = map.filter(s => {
+      if (categoriesFiltered === categoryCount) {
+        return true
+      }
+
+      categoriesFiltered++
+
+      return s !== categoryMapSymbol
+    }).filter((_, i) => i !== restIndex)
```
andreldsa: Try to write meaningful variable names.

react/constants/SearchHelpers.js
```diff
@@ -93,28 +93,42 @@ function removeFilter(map, rest, { type, slug, pagesPath }) {
...
+
+    const lengthDifference = map.length - mapWithoutCategories.length - 1
```
andreldsa: What if this result becomes negative?

lucasecdb: It never will.

The `mapWithoutCategories` is at least 1 unit smaller than the `map` array because I'm removing one filter from the array.

react/constants/SearchHelpers.js
lucasecdb: But `i` stands for index.

PR #45, opened 2018-09-20
Fix breadcrumb being shown on mobile

lucasecdb:
Add `dn-s` class to breadcrumb wrapper so it's hidden on mobile, also remove the content loader of the filters on mobile.

The breadcrumb isn't supposed to be displayed on mobile, and it was breaking the grid layout.

[Workspace](https://lucas--storecomponents.myvtex.com/Eletronicos/Smartphones), access with a espertofone or with mobile simulation.

#### Types of changes

### DIFF
--- a/react/components/FiltersContainer.js
+++ b/react/components/FiltersContainer.js
@@ -1,8 +1,8 @@
-/* global __RUNTIME__ */
 import React, { Component, Fragment } from 'react'
 import PropTypes from 'prop-types'
 import { flatten, path, identity, contains } from 'ramda'
 import ContentLoader from 'react-content-loader'
+import { withRuntimeContext } from 'render'
 
 import SelectedFilters from './SelectedFilters'
 import AvailableFilters from './AvailableFilters'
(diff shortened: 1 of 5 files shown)


react/components/SearchResult.js
```diff
@@ -38,7 +38,7 @@ export default class SearchResult extends Component {
 
     return (
       <div className="vtex-search-result vtex-page-padding pv5 ph9-l ph7-m ph5-s">
-        <div className="vtex-search-result__breadcrumb">
+        <div className="vtex-search-result__breadcrumb db-m dn-s">
```
estacioneto: why not `db-ns` ? More intuitive, don't you think?

lucasecdb: nice

react/components/FiltersContainer.js
```diff
@@ -123,7 +123,9 @@ export default class FiltersContainer extends Component {
       hiddenFacets,
     } = this.props
 
-    if (loading) {
+    const mobile = __RUNTIME__.hints.mobile
```
claudivanfilho: use withRuntimeContext

lucasecdb: This isn't included in the runtime context.

claudivanfilho: ok :)

PR #57, opened 2018-10-30
Fix slug of facet types other than brand being modified

lucasecdb:
Remove regex on slug and apply it only on brand slug's, e.g. `"Make B." -> "Make-B-"`

The regex should only be applied on slugs of the brands facets, the name is fine for the other types.

Access the [workspace](https://fixslugs--storecomponents.myvtex.com/eletronicos/smartphones?map=c%2Cc) and select any filter under the "Operational System".

#### Types of changes

### DIFF
--- a/react/constants/SearchHelpers.js
+++ b/react/constants/SearchHelpers.js
@@ -247,7 +247,14 @@ export function mountOptions(options, type, map, rest) {
   const restMap = restMapped(rest, map)
 
   return options.reduce((acc, opt) => {
-    const slug = opt.Slug || (opt.normalizedName || opt.Name).replace(/[^\w\d]/g, '-')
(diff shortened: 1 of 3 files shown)


react/constants/SearchHelpers.js
```diff
@@ -247,7 +247,14 @@ export function mountOptions(options, type, map, rest) {
   const restMap = restMapped(rest, map)
 
   return options.reduce((acc, opt) => {
-    const slug = opt.Slug || (opt.normalizedName || opt.Name).replace(/[^\w\d]/g, '-')
+    let slug
+
+    if (opt.type === BRANDS_TYPE) {
```
brunojdo: We need to put this in store-graphql

PR #59, opened 2018-10-31
Use icons from dreamstore icon pack

lucasecdb:
Use icons from dreamstore icon pack

[Workspace](https://lucas--storecomponents.myvtex.com/eletronicos/d?map=c%2Cc&rest=Smartphones).

#### Types of changes

* [x] Breaking change (fix or feature that would cause existing functionality to change)


brunojdo:
@matheusps change to your approach.

matheusps:
> @matheusps change to your approach.

Done ✅. You can access through this [workspace](https://matheusps--storecomponents.myvtex.com/iPho/s?map=ft)

react/images/Arrow.js
```diff
@@ -1,42 +1,41 @@
...
+      xmlnsXlink="http://www.w3.org/1999/xlink"
+      width={width}
+      height={height}
+      viewBox="0 0 14 10"
+      fill="none"
+    >
+      <Use id={iconId} />
+    </svg>
+  )
+}
```
vitoriaHeliane: Add a white line here.

react/images/InlineGrid.js
```diff
@@ -1,20 +1,21 @@
 import React from 'react'
 import PropTypes from 'prop-types'
+import Use from 'vtex.react-use-svg/Use'
 
 export default function InlineGrid({ active }) {
   const color = active ? '#828282' : '#ececec'
```
vitoriaHeliane: You can use Tachyons here if add a `div` parent and set its color by using tachyons' classes.

react/images/Grid.js
```diff
@@ -1,19 +1,21 @@
 import React from 'react'
 import PropTypes from 'prop-types'
+import Use from 'vtex.react-use-svg/Use'
 
 export default function Grid({ active }) {
   const color = active ? '#828282' : '#ececec'
```
vitoriaHeliane: You can use Tachyons here if add a `div` parent and set its color by using tachyons' classes.

react/images/Arrow.js
vitoriaHeliane: Add a white line here.

react/images/Grid.js
vitoriaHeliane: You can use Tachyons here if add a `div` parent and set its color by using tachyons' classes.

react/images/Arrow.js
vitoriaHeliane: Add a white line here.

react/images/Grid.js
vitoriaHeliane: You can use Tachyons here if add a `div` parent and set its color by using tachyons' classes.

manifest.json
```diff
@@ -28,6 +28,7 @@
     "vtex.styleguide": "6.x",
     "vtex.product-summary": "1.x",
     "vtex.store-graphql": "2.x",
-    "vtex.store": "1.x"
+    "vtex.store": "1.x",
+    "vtex.react-use-svg": "1.x"
```
brunojdo: 0.x

react/images/Check.js
```diff
@@ -1,26 +1,18 @@
 import React from 'react'
 import PropTypes from 'prop-types'
+import Use from 'vtex.react-use-svg/Use'
 
 export default function Check({ size }) {
   return (
     <svg
+      xmlns="http://www.w3.org/2000/svg"
```
brunojdo: This is always the same, why you don't put in Use?

matheusps: I think puting all of this in a single component would be a little strange. An idea that came in mind is to create another component called Svg, that wraps svg tags with default values for props xmlns and xmlnsXlink. Something like:

```
...
const Svg = ({ children, ...props}) => (
  <svg {...props}>
    { children }
  </svg>
);

Svg.defaultProps = {
  xmlns: 'http://www.w3.org/2000/svg',
  xmlnsXlink='http://www.w3.org/1999/xlink'
};
```

Using the component would be like: 
```
...
<Svg 
  width={size}
  height={size}
  viewBox="0 0 9 9"
  fill="none"
>
  <Use id="check" />
</Svg>
```

This Component would be useful in any other place that we use svgs. What do you think ?

manifest.json
matheusps: Fixed

react/images/Arrow.js
matheusps: Fixed

react/images/Grid.js
matheusps: Done

react/images/CheckTick.js
```diff
@@ -1,21 +1,20 @@
...
-      </svg>
-    )
-  }
+
+const CheckTick = () => {
+  return (
+    <svg
+      xmlns="http://www.w3.org/2000/svg"
+      xmlnsXlink="http://www.w3.org/1999/xlink"
+      width="11"
```
vitoriaHeliane: I think that the icons size should be able to change by props.

react/images/GridIcon.js
```diff
@@ -0,0 +1,38 @@
...
+
+GridIcon.propTypes = {
+  /** Whether the icon is active */
+  active: bool,
+  activeClassName: string,
+  muttedClassName: string,
+  width: string,
+  height: string,
+  fill: string,
+  iconID: string.isRequired
```
lucasecdb: rename to `icon` or `type`

PR #68, opened 2018-11-21
Scroll to top of search result when there's any change

athoscouto:
Scroll the window to the top of the search result whenever it changes.

When search result shrinks user end up seeing only the footer of the page.


athoscouto:
Addressed comments. Feel free to test it at https://athos--boticario.myvtex.com/perfumaria/d

PR #85, opened 2018-12-19
Use react virtualized

rerissondaniel:

Use [React Virtualized](https://bvaughn.github.io/react-virtualized/#/components/List) to improve the performance of the gallery rendering.

Instead of rendering all items, now the gallery only renders the visible items and pre-renders the next items to be shown.

[workspace](https://rerissontst--storecomponents.myvtex.com/clothing)
#### Types of changes

### DIFF
--- a/react/Gallery.js
+++ b/react/Gallery.js
@@ -1,7 +1,7 @@
 import React from 'react'
 import classNames from 'classnames'
 import { List, WindowScroller, AutoSizer, CellMeasurer, CellMeasurerCache } from 'react-virtualized'
-import { PropTypes } from 'prop-types'
+import PropTypes from 'prop-types'
 import { Adopt } from 'react-adopt'
 
 import { withRuntimeContext, NoSSR } from 'render'
@@ -34,7 +34,7 @@ const Gallery = ({
 
   cache.clearAll()
 
-  const toJsxItem = (item) => (
+  const renderItem = (item) => (
     <div
       key={item.productId}
       className="vtex-gallery__item mv2 pa1"
@@ -65,7 +65,7 @@ const Gallery = ({
       >
         {({ measure }) => (
           <div className={containerClasses} key={key} style={style} onLoad={measure}>
-            {rowItems.map(toJsxItem)}
+            {rowItems.map(renderItem)}
           </div>
         )}
       </CellMeasurer>
@@ -78,7 +78,7 @@ const Gallery = ({
 
   const ssrFallBack = (
     <div className={ssrContainer}>
-      {products.map(toJsxItem)}
+      {products.map(renderItem)}
     </div>
   )
 
--- a/react/Gallery.js
+++ b/react/Gallery.js
@@ -34,7 +34,7 @@ const Gallery = ({
 
   cache.clearAll()
 
-  const renderItem = (item) => (
+  const renderItem = item => (
     <div
       key={item.productId}
       className="vtex-gallery__item mv2 pa1"
@@ -86,7 +86,6 @@ const Gallery = ({
     <NoSSR onSSR={ssrFallBack}>
       <div className="vtex-gallery pa3 bn">
         <Adopt
-          key={layoutMode}
           mapper={{
             scroller: <WindowScroller />,
             autoSizer: <AutoSizer disableHeight />,
--- a/react/Gallery.js
+++ b/react/Gallery.js
@@ -18,8 +18,9 @@ const Gallery = ({
   layoutMode,
   runtime: { hints: { mobile } },
   itemWidth }) => {
+  const TWO_ITEMS = 2
+
   const cache = new CellMeasurerCache({
-    defaultHeight: 305,
     fixedWidth: true,
     keyMapper: () => 1,
   })
@@ -28,7 +29,7 @@ const Gallery = ({
     const from = index * itemsPerRow
     const rowItems = products.slice(from, from + itemsPerRow)
 
-    const containerClasses = classNames('Row vtex-gallery__row', {
+    const containerClasses = classNames('vtex-gallery__row', {
       'vtex-gallery__row--two-columns': layoutMode === 'small' && mobile,
     })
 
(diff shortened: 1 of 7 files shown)


react/Gallery.js
```diff
@@ -1,34 +1,104 @@
...
-    'vtex-gallery--two-columns': layoutMode === 'small',
+
+const Gallery = ({
+  products,
+  summary,
+  layoutMode,
+  runtime: { hints: { mobile } },
+  itemWidth }) => {
+  const cache = new CellMeasurerCache({
+    defaultHeight: 305,
```
luciannojunior: Fixed value?

react/Gallery.js
```diff
@@ -1,34 +1,104 @@
...
-        )
-      })}
+    <div className="vtex-gallery pa3 bn">
+      <NoSSR onSSR={renderOnServer()}>
+        <WindowScroller>
+          {({ height }) => {
+            return (
+              <AutoSizer disableHeight>
+                {({ width }) => {
+                  const itemsPerRow = (layoutMode === 'small' && mobile) ? 2 : (Math.floor(width / itemWidth) || 1)
```
luciannojunior: Extract the 2 into a constant

react/Gallery.js
```diff
@@ -1,34 +1,104 @@
...
+    defaultHeight: 305,
+    fixedWidth: true,
+    keyMapper: () => 1,
   })
 
+  const renderRow = ({ index, key, style, parent, itemsPerRow }) => {
+    const from = index * itemsPerRow
+    const rowItems = products.slice(from, from + itemsPerRow)
+
+    const containerClasses = classNames('Row vtex-gallery__row', {
```
luciannojunior: Why this class `Row`?

react/Gallery.js
rerissondaniel: Its an estimate to the initial unmeasured rows Height, but its not required

react/Gallery.js
rerissondaniel: I thought it was needed to react virtualized identify the rows, but it isn't, thanks for asking.

react/Gallery.js
```diff
@@ -1,34 +1,105 @@
...
-        )
-      })}
+    <div className="vtex-gallery pa3 bn">
+      <NoSSR onSSR={renderOnServer()}>
+        <WindowScroller>
+          {({ height }) => {
+            return (
+              <AutoSizer disableHeight>
+                {({ width }) => {
+                  const itemsPerRow = (layoutMode === 'small' && mobile) ? TWO_ITEMS : (Math.floor(width / itemWidth) || 1)
```
luciannojunior: Add ONE_ITEM. Makes sense

react/Gallery.js
```diff
@@ -1,34 +1,105 @@
...
+  products,
+  summary,
+  layoutMode,
+  runtime: { hints: { mobile } },
+  itemWidth }) => {
+  const TWO_ITEMS = 2
+
+  const cache = new CellMeasurerCache({
+    fixedWidth: true,
+    keyMapper: () => 1,
```
luciannojunior: this 1 is for itens too?

react/global.css
```diff
@@ -133,6 +133,16 @@
   margin: 0 auto;
 }
 
+.vtex-gallery__row {
+  display: grid;
+  grid-template-columns: repeat(auto-fill, minmax(280px, 1fr));
```
luciannojunior: cant use rem?

rerissondaniel: Done, I didn't think could be used here, thanks

react/Gallery.js
rerissondaniel: Yes, but in this case I think that using named constants doesn't make too much sense. I'll add anyway to maintain consistency.

react/Gallery.js
```diff
@@ -1,34 +1,106 @@
...
+const Gallery = ({
+  products,
+  summary,
+  layoutMode,
+  runtime: { hints: { mobile } },
+  itemWidth }) => {
+  const TWO_ITEMS = 2
+  const ONE_ITEM = 1
+
+  const cache = new CellMeasurerCache({
```
lucasecdb: move out of render

react/Gallery.js
```diff
@@ -1,34 +1,106 @@
...
-            />
-          </div>
-        )
-      })}
+    <div className="vtex-gallery pa3 bn">
+      <NoSSR onSSR={renderOnServer()}>
+        <WindowScroller>
+          {({ height }) => {
+            return (
+              <AutoSizer disableHeight>
```
lucasecdb: use `Adapt` with this and the `WindowScroller` component

react/Gallery.js
```diff
@@ -1,34 +1,106 @@
...
+                  return (<List
+                    key={layoutMode}
+                    deferredMeasurementCache={cache}
+                    autoHeight
+                    width={width}
+                    height={height}
+                    rowCount={nRows}
+                    rowHeight={cache.rowHeight}
+                    rowRenderer={args => renderRow({ ...args, itemsPerRow })}
+                  />)
```
lucasecdb: move the `)` to the line below :pick:

react/Gallery.js
```diff
@@ -1,36 +1,119 @@
 import React from 'react'
-import { PropTypes } from 'prop-types'
 import classNames from 'classnames'
+import { List, WindowScroller, AutoSizer, CellMeasurer, CellMeasurerCache } from 'react-virtualized'
+import { PropTypes } from 'prop-types'
```
estacioneto: Are you sure it shouldn't be `import PropTypes from 'prop-types'` ?

rerissondaniel: It was already there, thanks for catching it!

react/Gallery.js
```diff
@@ -1,36 +1,119 @@
...
+    <div className={ssrContainer}>
+      {products.map(renderItem)}
     </div>
   )
+
+  return (
+    <NoSSR onSSR={ssrFallBack}>
+      <div className="vtex-gallery pa3 bn">
+        <Adopt
+          key={layoutMode}
```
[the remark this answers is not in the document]
rerissondaniel: No, I had added it trying to solve a big, but it wasn't related to that, thanks

PR #96, opened 2019-01-28
Bar appearing on scroll up and disappearing on scroll down - mobile view

paul0vinicius:
Makes the search options bar always visible on mobile view but also taking advantage of the whole screen.

The fact that the user may want to filter the products while he is seeing them, so instead of scroll all the way up, he may just scroll a little bit and have the options right there.

[Workspace Link](https://searchresultoptionsbar--storecomponents.myvtex.com/clothing) in mobile view and scroll down to disappear and up to appear.

captura de tela de 2019-01-28 09-48-00

#### Types of changes

### DIFF
--- a/react/components/SearchResult.js
+++ b/react/components/SearchResult.js
@@ -241,7 +241,7 @@ class SearchResult extends Component {
 
     const showLoading = loading && !fetchMoreLoading
     const showContentLoader = showLoading && !showLoadingAsOverlay
-    const searchOptionsBarClasses = classNames({ 'flex justify-center flex-auto fixed z-1 w-100 bg-base bb bw1 b--light-gray pr4': mobile })
+    const searchOptionsBarClasses = classNames({ 'flex justify-center flex-auto fixed z-1 w-100 mw9 bg-base bb bw1 b--light-gray': mobile })
 
     return (
       <LoadingOverlay loading={showLoading && showLoadingAsOverlay}>
--- a/react/components/SearchResult.js
+++ b/react/components/SearchResult.js
@@ -1,18 +1,17 @@
 import React, { Component } from 'react'
 import { Spinner } from 'vtex.styleguide'
 import { ExtensionPoint, withRuntimeContext } from 'vtex.render-runtime'
-
+import classNames from 'classnames'
 import LoadingOverlay from './LoadingOverlay'
 import { searchResultPropTypes } from '../constants/propTypes'
 import LayoutModeSwitcher, { LAYOUT_MODE } from './LayoutModeSwitcher'
 
 import searchResult from '../searchResult.css'
 
-const HEADER_SIZE = 36
-const CONVERSION_MARGIN_VALUE = 0.08
-const MARGIN_TOP = "-2.4rem"
+const MARGIN_TOP_VISIBLE_SEARCH_BAR_OPTIONS = "0rem"
+const MARGIN_TOP_INVISIBLE_SEARCH_BAR_OPTIONS = "-5.5rem"
 const FADE_DURATION = "500ms" // It should be passed via props in the future
-const FADE_DELAY = "0ms" // Same as above
+const FADE_DELAY = "80ms" // Same as above
 
 /**
  * Search Result Component.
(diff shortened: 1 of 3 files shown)


react/components/SearchResult.js
```diff
@@ -101,6 +110,113 @@ class SearchResult extends Component {
...
+      if (scrollValue < HEADER_SIZE) searchOptionsBarElement.style.marginTop = `-${marginValue}rem`
+      else searchOptionsBarElement.style.marginTop = `${MARGIN_TOP}rem`
+    }
+  }
+
+  handleScrollDown = () => {
+    const searchOptionsBarElement = this.searchOptionsBar.current
+    if (searchOptionsBarElement) {
+      searchOptionsBarElement.style.opacity = 0
+      searchOptionsBarElement.style.transition = `opacity ${FADE_DURATION}ms ${FADE_DELAY}ms`
```
rerissondaniel: I think that maybe it was better to reduce the height of the bar, giving the impression that it was being scrolled up, instead of setting the opacity

react/searchResult.css
```diff
@@ -237,6 +240,17 @@
...
+  }
+
+  .searchOptionsBar{
+    display: grid;
+    display: -ms-grid;
+    -ms-grid-columns: 2fr 1px 1.5fr 1px 0.2fr;
+    grid-template-columns: 2fr 1px 1.5fr 1px 0.2fr;
+    grid-template-areas:
+      "orderby border filters borderseq switch";
+    width: 96%
```
rerissondaniel: Why not 100%? Its akward to see the shadows of elements that are behind the bar (like the shadow of the footer)

react/components/SearchResult.js
```diff
@@ -101,6 +110,113 @@ class SearchResult extends Component {
...
+    document.removeEventListener('scroll', this.handleScroll)
+  }
+
+  handleScroll = () => {
+    const scroll = this.props.leanMode ? Infinity : window.scrollY
+    const {runtime: { hints: { mobile } }} = this.props
+    
+    if (typeof scroll !== 'number' || !mobile) return
+
+    if (scroll > this.state.scrollValue) this.handleScrollDown()
```
theomoura: The `scroll` can be equal to  `this.state.scrollValue`? if yes, what happens in this situation? If not, you can use ternary conditional

react/components/SearchResult.js
```diff
@@ -10,6 +10,12 @@ import LayoutModeSwitcher, { LAYOUT_MODE } from './LayoutModeSwitcher'
 
 import searchResult from '../searchResult.css'
 
+const HEADER_SIZE = 36
+const CONVERSION_MARGIN_VALUE = 0.08
+const MARGIN_TOP = -2.4 //rem
```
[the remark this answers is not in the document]
paul0vinicius: Makes sense. I made a change.

react/components/SearchResult.js
```diff
@@ -101,6 +110,113 @@ class SearchResult extends Component {
...
+
+  handleScrollUp = scrollValue => {
+    const marginValue = scrollValue * CONVERSION_MARGIN_VALUE
+
+    const searchOptionsBarElement = this.searchOptionsBar.current
+    if (searchOptionsBarElement) {
+      searchOptionsBarElement.style.transition = `opacity ${FADE_DURATION}ms ${FADE_DELAY}ms`
+      searchOptionsBarElement.style.opacity = 1
+
+      if (scrollValue < HEADER_SIZE) searchOptionsBarElement.style.marginTop = `-${marginValue}rem`
```
theomoura: You could use ternary conditional here.

react/components/SearchResult.js
```diff
@@ -101,6 +110,113 @@ class SearchResult extends Component {
...
+
+    this.setState({scrollValue: scroll})
+  }
+
+  handleScrollUp = scrollValue => {
+    const marginValue = scrollValue * CONVERSION_MARGIN_VALUE
+
+    const searchOptionsBarElement = this.searchOptionsBar.current
+    if (searchOptionsBarElement) {
+      searchOptionsBarElement.style.transition = `opacity ${FADE_DURATION} ${FADE_DELAY}`
```
Klynger: You should apply this transition to `WebkitTransition` too in order support older browsers, so you should have these two lines:
```suggestion
searchOptionsBarElement.style.transition = `opacity ${FADE_DURATION} ${FADE_DELAY}`
searchOptionsBarElement.style.WebkitTransition = `opacity ${FADE_DURATION} ${FADE_DELAY}`
```

react/components/SearchResult.js
```diff
@@ -101,6 +110,113 @@ class SearchResult extends Component {
...
+      if (scrollValue < HEADER_SIZE) searchOptionsBarElement.style.marginTop = `-${marginValue}rem`
+      else searchOptionsBarElement.style.marginTop = MARGIN_TOP
+    }
+  }
+
+  handleScrollDown = () => {
+    const searchOptionsBarElement = this.searchOptionsBar.current
+    if (searchOptionsBarElement) {
+      searchOptionsBarElement.style.opacity = 0
+      searchOptionsBarElement.style.transition = `opacity ${FADE_DURATION} ${FADE_DELAY}`
```
Klynger: Same thing here

react/components/SearchResult.js
```diff
@@ -10,6 +10,12 @@ import LayoutModeSwitcher, { LAYOUT_MODE } from './LayoutModeSwitcher'
 
 import searchResult from '../searchResult.css'
 
+const HEADER_SIZE = 36
+const CONVERSION_MARGIN_VALUE = 0.08
+const MARGIN_TOP = "-2.4rem"
```
Klynger: Use single-quoted string

react/components/SearchResult.js
```diff
@@ -101,6 +110,113 @@ class SearchResult extends Component {
...
+      searchOptionsBarElement.style.opacity = 1
+
+      if (scrollValue < HEADER_SIZE) searchOptionsBarElement.style.marginTop = `-${marginValue}rem`
+      else searchOptionsBarElement.style.marginTop = `${MARGIN_TOP}rem`
+    }
+  }
+
+  handleScrollDown = () => {
+    const searchOptionsBarElement = this.searchOptionsBar.current
+    if (searchOptionsBarElement) {
```
theomoura: There is any cenario where `searchOptionsBarElement` is null or undefined?

react/components/SearchResult.js
```diff
@@ -101,6 +110,113 @@ class SearchResult extends Component {
...
+      specificationFilters,
+      tree,
+      hiddenFacets,
+    } = this.state
+
+    const hideFacets = !map || !map.length
+    const filterClasses = classNames({ 'flex justify-center flex-auto': mobile })
+
+    return (
+      <React.Fragment>
```
theomoura: You just need to use `<Fragment>`

react/components/SearchResult.js
paul0vinicius: I have changed the direct style application for a class application based on the state change. The animations are from animate.css library and I think it covers this case.

react/components/SearchResult.js
paul0vinicius: Done!

react/components/SearchResult.js
paul0vinicius: Done!

react/components/SearchResult.js
paul0vinicius: I am not sure. In certains examples I have seen this is a safe way to access the DOM element, but I think it is not a requirement. Anyway, this function is not used anymore because the strategy of appearing/disappearing has changed.

PR #97, opened 2019-01-29
Normalized product sort skus instead of the sum of skus `AvailableQuantity`

JuanAlmeida12:

As the title says. This is a better approach. 

[Access this workspace](https://returnurl--storecomponents.myvtex.com)

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+## [3.10.3] - 2019-01-30
 ### Fixed
 - Normalized product sorts SKU's instead of summing their `AvailableQuantity`.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 ### Fixed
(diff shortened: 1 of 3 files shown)


react/components/GalleryItem.js
```diff
@@ -22,21 +22,17 @@ export default class GalleryItem extends Component {
     if (!product) {
       return null
     }
+    const compareSKUs = (skuA, skuB) => path(['sellers', '0', 'commertialOffer', 'AvailableQuantity'], skuB) - path(['sellers', '0', 'commertialOffer', 'AvailableQuantity'], skuA)
```
luciannojunior: Extract the `path(['sellers', '0', 'commertialOffer', 'AvailableQuantity']` into a function and call it

luciannojunior: And use [comparator](https://ramdajs.com/docs/#comparator) with [subtract](
https://ramdajs.com/docs/#subtract)

luciannojunior: Avoid naming parameters when you can omit them. Ramda uses curry for every function they export, as I've pointed in the last PR

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- Normalized product sort skus instead of the sum of skus `AvailableQuantity`.
```
luciannojunior: _sort SKU's instead of summing their `AvailableQuantity`_

react/components/GalleryItem.js
luciannojunior: And make it clean in the name of the comparator which criteria is being used.

PR #101, opened 2019-01-30
Fix Gallery Scroll and Query

rerissondaniel:

Fix the query that loaded more items and fix the size of the gallery items when more items were loaded.

### What problem is this solving?
The loadmore wasn't working at all and the layout broke when more items were loaded.

[workspace](https://fixsrscroll--phmex.myvtex.com/pizza)

#### Types of changes

### DIFF
--- a/react/Gallery.js
+++ b/react/Gallery.js
@@ -34,6 +34,7 @@ const Gallery = ({
   itemWidth,
 }) => {
 
+  //Necessary to compute layout properly (e.g. from 'inline' to 'normal' or small)
   cache.clearAll()
 
   const renderItem = item => (
@@ -92,17 +93,19 @@ const Gallery = ({
             scroller: <WindowScroller />,
             autoSizer: <AutoSizer disableHeight />,
           }}
-          mapProps={({ scroller: { height }, autoSizer: { width } }) => ({ width, height })}
+          mapProps={({ scroller: { height, scrollTop }, autoSizer: { width } }) => ({ width, height, scrollTop })}
         >
-          {({ width, height }) => {
+          {({ width, height, scrollTop }) => {
             const itemsPerRow = (layoutMode === 'small' && mobile) ? TWO_ITEMS : (Math.floor(width / itemWidth) || ONE_ITEM)
             const nRows = Math.ceil(products.length / itemsPerRow)
-
+          
             return (
               <List
                 key={layoutMode}
                 deferredMeasurementCache={cache}
+                overscanRowCount={2}
                 autoHeight
+                scrollTop={scrollTop}
                 width={width}
                 height={height}
                 rowCount={nRows}
--- a/react/Gallery.js
+++ b/react/Gallery.js
@@ -3,6 +3,7 @@ import classNames from 'classnames'
 import { List, WindowScroller, AutoSizer, CellMeasurer, CellMeasurerCache } from 'react-virtualized'
 import PropTypes from 'prop-types'
 import { Adopt } from 'react-adopt'
+import { map } from 'ramda'
 
 import { withRuntimeContext, NoSSR } from 'vtex.render-runtime'
 
@@ -69,7 +70,7 @@ class Gallery extends Component {
         rowIndex={index}
       >
         <div className={containerClasses} key={key} style={style}>
-          {rowItems.map(this.renderItem)}
+          {map(this.renderItem, rowItems)}
         </div>
       </CellMeasurer>
     )
@@ -84,9 +85,14 @@ class Gallery extends Component {
 
     return (
       <div className={ssrContainer}>
-        {products.map(this.renderItem)}
+        {map(this.renderItem, products)}
       </div>
     )
+  }
+
+  updateCache = layoutMode => {
+      this.setState({ layoutMode })
+      this.cache.clearAll()
   }
 
   render() {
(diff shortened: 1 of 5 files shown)


react/components/SearchResultContainer.js
```diff
@@ -128,9 +129,9 @@ export default class SearchResultContainer extends Component {
 
     this._fetchMoreLocked = true
 
-    const { maxItemsPerPage, searchQuery: { products } } = this.props
+    const { maxItemsPerPage, searchQuery: { products, recordsFiltered } } = this.props
 
-    const to = maxItemsPerPage + products.length - 1
+    const to = Math.min(+maxItemsPerPage + products.length, recordsFiltered) - 1
```
JuanAlmeida12: You can use ramda functions to this, like composing `subtract`  with `min` and `sum`. What you think?

react/Gallery.js
```diff
@@ -59,62 +62,81 @@ const Gallery = ({
...
-      </div >
-    </NoSSR>
-  )
+    const ssrContainer = classNames(`${searchResult.gallery} pa3 bn`, {
+      [searchResult.galleryTwoColumns]: layoutMode === 'small' && mobile,
+    })
+
+    return (
+      <div className={ssrContainer}>
+        {products.map(this.renderItem)}
```
JuanAlmeida12: You can use ramda `map` here.

react/Gallery.js
```diff
@@ -59,62 +62,81 @@ const Gallery = ({
...
+        <div className={`${searchResult.gallery} pa3 bn`}>
+          <Adopt
+            mapper={{
+              scroller: <WindowScroller />,
+              autoSizer: <AutoSizer disableHeight />,
+            }}
+            mapProps={({ scroller: { height, scrollTop, isScrolling }, autoSizer: { width } }) => ({ width, height, scrollTop, isScrolling })}
+          >
+            {({ width, height, scrollTop, isScrolling }) => {
+              const itemsPerRow = (layoutMode === 'small' && mobile) ? TWO_ITEMS : (Math.floor(width / itemWidth) || ONE_ITEM)
```
theomoura: You could move this to outside `return` block for better readability, what do you think?

react/Gallery.js
```diff
@@ -59,62 +62,81 @@ const Gallery = ({
...
+
+    return (
+      <NoSSR onSSR={this.handleSSR()}>
+        <div className={`${searchResult.gallery} pa3 bn`}>
+          <Adopt
+            mapper={{
+              scroller: <WindowScroller />,
+              autoSizer: <AutoSizer disableHeight />,
+            }}
+            mapProps={({ scroller: { height, scrollTop, isScrolling }, autoSizer: { width } }) => ({ width, height, scrollTop, isScrolling })}
```
theomoura: You could put `mapProps` value on a constant outside return block for better readability, what do you think?

react/Gallery.js
```diff
@@ -59,62 +62,81 @@ const Gallery = ({
...
+          >
+            {({ width, height, scrollTop, isScrolling }) => {
+              const itemsPerRow = (layoutMode === 'small' && mobile) ? TWO_ITEMS : (Math.floor(width / itemWidth) || ONE_ITEM)
+              const nRows = Math.ceil(products.length / itemsPerRow)
```
theomoura: The same here

react/Gallery.js
rerissondaniel: Yes, thanks for the suggestion.

react/Gallery.js
rerissondaniel: Render would be like that:
```js
render() {
    const {
      layoutMode,
    } = this.props

    // Reset the cache to recalculate the heights when layout changes
    if (layoutMode !== this.state.prevLayoutMode) {
      this.setState({ layoutMode })
      this.cache.clearAll()
    }

    const mapContainerProps = ({ scroller: { height, scrollTop, isScrolling }, autoSizer: { width } }) => ({ width, height, scrollTop, isScrolling })

    return (
      <NoSSR onSSR={this.handleSSR()}>
        <div className={`${searchResult.gallery} pa3 bn`}>
          <Adopt
            mapper={{
              scroller: <WindowScroller />,
              autoSizer: <AutoSizer disableHeight />,
            }}
            mapProps={mapContainerProps}
          >
            {this.renderList}
          </Adopt>
        </div >
      </NoSSR>
    )
  }
```
I think that doing it in that way separates related concepts and makes the render props harder to see. Also, I would be repeating some for for destructuring in renderList.

react/components/SearchResultContainer.js
rerissondaniel:
```js
const to = subtract(min(sum(maxItemsPerPage, products.length), recordsFiltered), 1)
```
It would be something like that, right?
I think that using the operators is more idiomatic.

react/Gallery.js
rerissondaniel: Thanks for the suggestion.

react/components/SearchResultContainer.js
theomoura: I prefer this solution anyway, but i will pass it and can be discussed forwardly

react/Gallery.js
theomoura: I prefer this solution, but anyway i will pass it and can be discussed forwardly

PR #104, opened 2019-02-04
Gallery spaces can be customized

JuanAlmeida12:

Spaces between `Gallery Items` can be customized now and theses spaces width won't change on screen resize.

[Access this workspace](https://galleryspace--storecomponents.myvtex.com)

galleryspace--storecomponents myvtex com_
galleryspace--storecomponents myvtex com_ 1

#### Types of changes

### DIFF
--- a/react/Gallery.js
+++ b/react/Gallery.js
@@ -37,8 +37,8 @@ class Gallery extends Component {
 
   renderItem = item => {
     const { summary, layoutMode, runtime: { hints: { mobile } } } = this.props
-    const galleryItemClasses = classNames(searchResult.galleryItem, {
-      'mv2 pa1': mobile
+    const galleryItemClasses = classNames(searchResult.galleryItem, 'pa1', {
+      'mv2': mobile
     })
     return (
       <div
@@ -66,7 +66,7 @@ class Gallery extends Component {
 
     const styleWithGap = {
       ...style,
-      gridGap: `${gap}rem`,
+      gridGap: gap,
     }
 
     return (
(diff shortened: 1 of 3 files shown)


react/searchResult.css
```diff
@@ -234,6 +231,10 @@
     grid-template-columns: 1fr;
   }
 
+  .galleryRow {
```
GustavoMA: Why are you defining `.galleryRow` class again? You could just add this to the one in line 93.

react/Gallery.js
```diff
@@ -61,6 +64,11 @@ class Gallery extends Component {
       [searchResult.galleryTwoColumns]: layoutMode === 'small' && mobile,
     })
 
+    const styleWithGap = {
+      ...style,
+      gridGap: gap,
```
theomoura: Check if `grid-gap` works on IE

react/Gallery.js
```diff
@@ -167,7 +177,8 @@ Gallery.propTypes = {
 Gallery.defaultProps = {
   maxItemsPerPage: 10,
   products: [],
-  itemWidth: 300,
+  itemWidth: 280,
```
theomoura: Why the item width was reduced here? should't it be 284 for the gap `16`?

react/searchResult.css
GustavoMA: Ops, it's inside a media query.

react/Gallery.js
luciannojunior: _Spoiler alert:_ it doesn't.

But so doesn't the grid anyway... By works I mean works

luciannojunior:
image
Did you check this?

luciannojunior:
Where is the `gap` being passed with value 32? I checked on your workspace

If it is right, shouldn't 32 be the default value?

JuanAlmeida12:
> Where is the `gap` being passed with value 32? I checked on your workspace
> 
> If it is right, shouldn't 32 be the default value?

I had changed in my workspace to show more space.

PR #109, opened 2019-02-07
Fix misuse of layoutMode prop in desktop view

GustavoMA:
As described in the title.

[Here](https://alves--phmex.myvtex.com/)

#### Types of changes


react/components/SearchResult.js
```diff
@@ -39,15 +39,15 @@ class SearchResult extends Component {
     hiddenFacets: this.props.hiddenFacets,
   }
 
-  handleLayoutChange = e => {
+  handleMobileLayoutChange = e => {
     e.preventDefault()
 
-    const defaultModes = [this.props.hiddenFacets.layoutMode1, this.props.hiddenFacets.layoutMode2]
-    const modeIndex = (defaultModes.indexOf(this.state.galleryLayoutMode) + 1) % 2
+    const defaultModes = [this.props.mobileLayout.mode1, this.props.mobileLayout.mode2]
```
rerissondaniel: I'm not sure if defaultModes makes sense here, maybe only 'modes' would be more descriptive and avoid misunderstandings

react/index.js
```diff
@@ -85,27 +85,34 @@ SearchResultQueryLoader.getSchema = props => {
     type: 'object',
     properties: {
       ...querySchema,
-      hiddenFacets: {
-        title: 'editor.search-result.hiddenFacets',
+      mobileLayout: {
+        title: 'Mobile Layout Switcher',
```
rerissondaniel: Ids should be used in order to support internationalization

react/components/SearchResult.js
GustavoMA: Fixed.

react/Gallery.js
```diff
@@ -54,8 +59,8 @@ Gallery.propTypes = {
   products: PropTypes.arrayOf(productShape),
   /** ProductSummary props. */
   summary: PropTypes.any,
-  /** Layout mode of the gallery */
-  layoutMode: PropTypes.string,
+  /** Layout mode of the gallery in mobile view */
+  mobileLayoutMode: PropTypes.string,
```
luciannojunior: Are the layout modes discrete? You could use PropTypes' `oneOf' to make it more clear. Maybe add some constantes on the top of the page and use it in the conditionals...

luciannojunior: There are some proptypes that are not required and algo they are not on defaultProps. Pls check

react/components/SearchResult.js
```diff
@@ -15,14 +15,14 @@ class SearchResult extends Component {
   static propTypes = searchResultPropTypes
 
   static defaultProps = {
-    hiddenFacets: {
-      layoutMode1: LAYOUT_MODE[0].value,
-      layoutMode2: LAYOUT_MODE[1].value,
+    mobileLayout: {
```
luciannojunior: Check if the [propTypes](https://github.com/vtex-apps/search-result/blob/master/react/constants/propTypes.js) are correct

react/index.js
```diff
@@ -85,27 +85,34 @@ SearchResultQueryLoader.getSchema = props => {
...
         type: 'object',
         isLayout: true,
         properties: {
-          layoutMode1: {
-            title: 'Layout Mode Switcher 1',
+          mode1: {
+            title: 'Layout Mode 1',
```
luciannojunior: Titles here could be internationalized

react/index.js
GustavoMA: Fixed.

react/index.js
GustavoMA: Fixed.

react/Gallery.js
GustavoMA: Fixed.

react/components/SearchResult.js
```diff
@@ -15,14 +15,14 @@ class SearchResult extends Component {
...
+      mode1: LAYOUT_MODE[0].value,
+      mode2: LAYOUT_MODE[1].value,
     },
   }
 
   state = {
-    galleryLayoutMode: this.props.hiddenFacets.layoutMode1 || LAYOUT_MODE[0].value,
+    mobileLayoutMode: this.props.mobileLayout.mode1,
```
theomoura: Please, double check this change, it was not working properly only with the props before.

GustavoMA: `this.props.mobileLayout.mode1` is already in the `defaultProps`.
But I also tested it out and it is working properly.

react/components/SearchResult.js
GustavoMA: Fixed.

PR #123, opened 2019-03-12
Fix outside click in mobile OrderBy dropdown.

JuanAlmeida12:

As the title says.

When clicking outside of the OrderBy dropdown, on mobile mode,  the box don't close.

[Access this workspace](https://clickoutside--storecomponents.myvtex.com/)

#### Types of changes

### DIFF
--- a/react/__mocks__/vtex.render-runtime.js
+++ b/react/__mocks__/vtex.render-runtime.js
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useState } from 'react'
 
 const config = { isMobile: false }
 
@@ -14,6 +14,11 @@ export const withRuntimeContext = Comp => {
 
 export const setMobileState = isMobile => {
   config.isMobile = isMobile
+}
+
+export const useRuntime = () => {
+  const [hints, setHints] = useState({ mobile: false, desktop: true })
+  return { hints }
 }
 
 export const ExtensionPoint = ({ id }) => <div> Extension Point: {id} </div>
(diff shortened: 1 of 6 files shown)


react/hooks/useOutsideClick.js
```diff
@@ -0,0 +1,13 @@
+import { useLayoutEffect } from 'react'
+
+const useOutsideClick = (ref, handler, when) => {
```
theomoura: Could you name `when` better? maybe `whenClicked`

react/__mocks__/vtex.render-runtime.js
```diff
@@ -16,6 +16,11 @@ export const setMobileState = isMobile => {
   config.isMobile = isMobile
 }
 
+export const useRuntime = () => {
+  const [hints, setHints] = useState({ mobile: false, desktop: true })
```
matheusps: `useState` here is unnecessary beacuse youŕe just exporting the object hints...

theomoura: since this is a mock of the runtime, I think useState makes sense

matheusps: Nope, actually. I'll break it down for you. What was beeing done here:
1. After line 20, your hints will have a initial value that you passed to useState, that is `{ mobile: false, desktop: true }`.
2. You do nothing with this value and return it.

Sooo, you will export your initial value EVERY TIME. And you cannot change it, since you do not return the `setHints` as well.
Finally, this code can be refactored to: 
`export const useRuntime = ( ) => { hints: { mobile: false, desktop: true} } `

Got it ?

PR #137, opened 2019-03-29
Added the department before each category to pass it to Breadcrumbs

matheusps:
As the title says.

**Fix**
*Should pass:* `[ '/Clothing', '/Clothing/Shirts' ]`
*Was passing:* `[ '/Clothing', '/Shirts' ]`

Linked @ [workspace](https://catorder--storecomponents.myvtex.com). ⚠️Test for more than one department/category

#### Types of changes

### DIFF
--- a/react/components/SearchResultContainer.js
+++ b/react/components/SearchResultContainer.js
@@ -39,8 +39,10 @@ export default class SearchResultContainer extends Component {
 
     const categoryWithChildrenReducer = (acc, category) => [
       ...acc,
-      category.Link,
-      ...category.Children.reduce(categoryReducer, []),
+      `/${category.Name}`,
+      ...category.Children.reduce(categoryReducer, []).map(
+        children => `/${category.Name}${children}`
+      ),
     ]
 
     const getCategoryList = (reducer, initial = []) =>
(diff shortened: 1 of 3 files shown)


react/components/SearchResultContainer.js
```diff
@@ -39,8 +39,10 @@ export default class SearchResultContainer extends Component {
 
     const categoryWithChildrenReducer = (acc, category) => [
       ...acc,
-      category.Link,
-      ...category.Children.reduce(categoryReducer, []),
+      `/${category.Name}`,
+      ...category.Children.reduce(categoryReducer, []).map(
+        children => `/${category.Name}${children}`
```
lbebber: generally what should be the result of this expression? can you give an example?

matheusps: Yep. The original array would be something like:
```
[
  { 
    name: categoryName,
    ...,
    Children: [  { name: childrenCategoryName, ... }, ... ] 
  }
  ...
]
```
The final result would be: `['\categoryName', '\categoryName\childrenCategoryName']`.
This matches the shape that `productQuery` returns the categories (on `product-details`).

lbebber: @matheusps ah, I see

 I just thought this bit was weird: `/${category.Name}${children}`, because there are no separators between `${category.Name}` and `${children}`. Is this intentional?

matheusps: Ooops, my fault. It was correct, but the reduce was unnecessary. Just made that simpler 😅. Thanks

PR #141, opened 2019-04-08
Add Category Panel Component

sahanljc:

Category panel is an optional representation of the category tree response returned by the VTEX product search API. It leverages the `CategoriesHighlights` from `vtex.store-components`. The visibility of the category panel defaults to `false`.

Stores often want to show the category images when a customer searches a product. `vtex.search-results` shows the available category filters in a left panel along with a link. It does not support showing an image.

1. Set the `showCategoryPanel` property of the `SearchResultQueryLoader` to `true`.
2. Search a product. (`quantityOfItemsPerRow` defaults to 4. You can optionally set it to 2.)

category-panel

#### Types of changes

* [x] Requires change to documentation, which has been updated accordingly.


react/CategoryPanel.js
```diff
@@ -0,0 +1,100 @@
...
+      }
+    )
+    let i = 0
+    const items = []
+    for (i = 0; i < categoriesHighlighted.length; i += quantityOfItemsPerRow) {
+      items.push(categoriesHighlighted.slice(i, i + quantityOfItemsPerRow))
+    }
+    return (
+      <Fragment key={`parent-fragment-${category.Id}`}>
+        <div className={headerClasses}>
```
klzns: I think there's no need for this div here, it could be the tag `<h3>` directly, since it's a block tag.

gabrielgalc:
Hi @sahanljc , thank you for the contribution!
Can you provide us a workspace so we can check the component's behavior? 
Thanks!

react/CategoryPanel.js
```diff
@@ -0,0 +1,100 @@
...
+    const categoriesHighlighted = category.Children.slice(0, noOfChildren).map(
+      child => ({
+        id: child.Id,
+        name: child.Name,
+        to: child.Link,
+      })
+    )
+    const headerClasses = classNames(
+      searchResult.categoryPanelHeaderRow,
+      'flex flex-row flex-wrap items-stretch pa3 bn',
```
klzns: I think it lacks a heading CSS token class here. Like `t-heading-4` or `5`.

You can read more about the typography classes here:
https://styleguide.vtex.com/#/Styles?id=typography

react/CategoryPanel.js
sahanljc: I thought to have that `<div>` just keep the HTML structure cleaner (the generated rows below the heading have `<div>`s in the same way) and have more control over the styling there.

sahanljc:
Hi @gabrielgalc, you can check the feature at [this](https://sahan--biscoindqa.myvtex.com) workspace. For detailed search results have a look at the [brand search for South Co](https://sahan--biscoindqa.myvtex.com/south-co).

You can do full text search for **4414 in Electronic Components**, **sample in Electronic Components**.

react/CategoryPanel.js
sahanljc: Thanks for pointing it out. I have added it at 1504d60.

sahanljc:
Do you guys think we should apply the map based filter in this panel as well? I am referring to the `getAvailableCategories` in `react/FilterNavigator.js`.

react/CategoryPanel.js
```diff
@@ -0,0 +1,100 @@
...
+      {
+        mh4: !mobile,
```
matheusps: I think you could just use `mh4-ns` tachyon instead of a conditional with the mobile from `runtime` 🤔

react/CategoryPanel.js
```diff
@@ -0,0 +1,100 @@
...
+      }
+    )
+    const itemClasses = classNames(
+      searchResult.categoryPanelItemRow,
+      'flex flex-row flex-wrap items-stretch pa3 bn',
+      {
+        mh4: !mobile,
```
matheusps: Here as well 🤓

react/CategoryPanel.js
```diff
@@ -0,0 +1,100 @@
+import React, { Component, Fragment } from 'react'
+import classNames from 'classnames'
+import PropTypes from 'prop-types'
+import { Link, withRuntimeContext } from 'vtex.render-runtime'
+import { facetOptionShape } from './constants/propTypes'
+import searchResult from './searchResult.css'
+import CategoriesHighlights from 'vtex.store-components/CategoriesHighlights'
+
+class CategoryPanel extends Component {
```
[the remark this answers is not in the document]
sahanljc: We plan to use a query here to get image URLs we store in Master Data. Will changing to a functional component allow us to have a query here? We have this change in vtex-apps/store-graphql#263. :smiley:

react/CategoryPanel.js
sahanljc: Good pick! Thanks! :sunglasses:

react/CategoryPanel.js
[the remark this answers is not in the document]
matheusps: Yep, you can still use `queries` with `apollo` even with a functional component.

react/CategoryPanel.js
sahanljc: Okay, I was late to figure this point out. It's done. Thanks man! :+1:

react/CategoryPanel.js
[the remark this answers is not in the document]
sahanljc: Okay, it's done. Thank you very much. :+1:

PR #145, opened 2019-04-12
Remove showTitle schema property

lucasecdb:
Remove the `showTitle` schema property.

The title in the search should be configured only in the blocks, not on the storefront.

#### Types of changes

### DIFF
--- a/react/Title.js
+++ b/react/Title.js
@@ -3,11 +3,7 @@ import React from 'react'
 
 import styles from './searchResult.css'
 
-const Title = ({ params, showTitle }) => {
-  if (!showTitle) {
-    return null
-  }
-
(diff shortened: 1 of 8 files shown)


react/Title.js
```diff
@@ -3,11 +3,7 @@ import React from 'react'
 
 import styles from './searchResult.css'
 
-const Title = ({ params, showTitle }) => {
-  if (!showTitle) {
-    return null
-  }
-
+const Title = ({ params }) => {
   const title = params.term ||
```
kevinch: No risk of confusion between `Title` and `title`?

lucasecdb: sure, I'll rename the component

PR #149, opened 2019-04-26
Fix undefined error on fetch more update

lucasecdb:
Read the query values directly, instead of reading from the `search` property.

image

[Workspace](https://lucas2--storecomponents.myvtex.com/teste).

#### Types of changes

### DIFF
--- a/react/components/SearchResultContainer.js
+++ b/react/components/SearchResultContainer.js
@@ -170,13 +170,11 @@ export default class SearchResultContainer extends Component {
         )
 
         return {
-          search: {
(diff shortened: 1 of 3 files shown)


react/components/SearchResultContainer.js
```diff
@@ -169,14 +169,25 @@ export default class SearchResultContainer extends Component {
           }
         )
 
+        // backwards compat
```
felippenardi: Avoiding abbreviations will avoid confusion in the future. I'd suggest using the full word `compatibility`

PR #150, opened 2019-04-26
Use stable filters

lucasecdb:
Add logic for selected filters in the components and use the facets value to add it to the url, instead of the `linkEncoded`, because it's not "reliable" anymore.

In order for the filters to be consistent with the products shown, we shouldn't fetch the facets more than once per category or full text (i.e. search). This is because if we change the filters for every change in the own filters, the quantity shown becomes inconsistent (because they are calculated on the filters sent to the api) and we lose the ability to filter by more than 1 brand, for example.

[This workspace](https://lucas--storecomponents.myvtex.com/clothing/shirts?map=c%2Cc), try changing the brand or specification filters.

#### Types of changes

### DIFF
--- a/react/FilterNavigator.js
+++ b/react/FilterNavigator.js
@@ -1,7 +1,7 @@
 import classNames from 'classnames'
 import PropTypes from 'prop-types'
 import { map, flatten, path, contains, filter, prop } from 'ramda'
-import React from 'react'
+import React, { useMemo } from 'react'
 import ContentLoader from 'react-content-loader'
 import { FormattedMessage } from 'react-intl'
 import { useRuntime } from 'vtex.render-runtime'
@@ -14,6 +14,7 @@ import {
   paramShape,
   hiddenFacetsSchema,
 } from './constants/propTypes'
+import useSelectedFilters from './hooks/useSelectedFilters'
 
 import searchResult from './searchResult.css'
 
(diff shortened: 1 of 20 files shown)


react/FilterNavigator.js
```diff
@@ -51,18 +52,20 @@ const FilterNavigator = ({
     hints: { mobile },
   } = useRuntime()
 
-  const getSelectedFilters = () => {
-    const availableCategories = filter(prop('selected'), getCategories(tree))
+  const selectedFilters = useSelectedFilters(
+    useMemo(() => {
```
lbebber: correct me if I'm wrong, because I have not looked too deeply into this, but shouldn't hooks be used only on the top level of the component?

react/components/FacetItem.js
```diff
@@ -21,6 +21,8 @@ const FacetItem = ({ facet }) => {
   const { query, map } = useContext(QueryContext)
 
   const handleChange = () => {
+    const urlParams = new URLSearchParams(window.location.search)
```
lbebber: could this ever be run on SSR? if so, shouldn't you handle the case where `window.location` doesn't exist?

react/FilterNavigator.js
lucasecdb: yes! that's a good point but the `useMemo` is actually on the top level, and I wanted to "skip" one variable declaration (to avoid naming things). I think it kind looks weird at the first glance, because one wouldn't expect for it to be like this, but it's totally valid.

react/components/FacetItem.js
lucasecdb: yes! because this is on the event handler

PR #156, opened 2019-05-06
Fix/local query

jgfidelis:
ver que most popular on home is working 
https://fidelis4--invictastores.myvtex.com/

#### Types of changes

### DIFF
--- a/react/components/LocalQuery.js
+++ b/react/components/LocalQuery.js
@@ -51,8 +51,8 @@ const LocalQuery = props => {
     from,
     to,
     withFacets: !!(
-      map &&
-      map.length > 0 &&
+      facetMap &&
+      facetMap.length > 0 &&
       facetQuery &&
       facetQuery.length > 0
     ),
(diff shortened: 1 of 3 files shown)


react/components/LocalQuery.js
```diff
@@ -29,23 +30,25 @@ const LocalQuery = props => {
   const from = (page - 1) * maxItemsPerPage
   const to = from + maxItemsPerPage - 1
 
+  const variables = {
+    query: queryField,
+    map,
```
lucasecdb: missing `facetMap` and `facetQuery`

react/components/LocalQuery.js
```diff
@@ -29,23 +30,40 @@ const LocalQuery = props => {
...
+    orderBy,
+    priceRange,
+    from,
+    to,
+    withFacets: !!(
+      map &&
+      map.length > 0 &&
```
lucasecdb: should be facetMap instead of map

react/components/LocalQuery.js
```diff
@@ -29,23 +30,40 @@ const LocalQuery = props => {
...
+      queryField &&
+      queryField.length > 0
```
lucasecdb: should be facetQuery instead of queryField

PR #169, opened 2019-05-10
Use recordsFiltered from facets query in not found page check

lucasecdb:
Update the check for when to display the not found page to use the recordsFiltered returned from the facets query, instead of the products.

This is so we can show the "not found" only in the gallery if the cause for the lack of products is an applied filter by the user, so they can unapply the filter and get the products back in the list.

[workspace](https://lucas--invictastores.myvtex.com/watches/1603/Technomarine?map=c%2CspecificationFilter_69%2Cc)

#### Types of changes

### DIFF
--- a/react/components/SearchResult.js
+++ b/react/components/SearchResult.js
@@ -28,6 +28,7 @@ class SearchResult extends Component {
     // on SSR the getDerivedStateFromProps isn't called
     products: this.props.products,
     recordsFiltered: this.props.recordsFiltered,
+    facetRecordsFiltered: this.props.facetRecordsFiltered,
     brands: this.props.brands,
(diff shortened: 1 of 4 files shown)


react/components/SearchResult.js
```diff
@@ -137,7 +141,11 @@ class SearchResult extends Component {
     const term =
       params && params.term ? decodeURIComponent(params.term) : undefined
 
-    if (recordsFiltered === 0 && !loading) {
+    // only show the not found page if the reason for it
+    // isn't because of some applied filter, but that we
+    // *really* don't have any products to show for the
+    // current query
```
luciannojunior: I think it's clear without the comment, but ok

PR #173, opened 2019-05-14
Send productClick events to PixelManager

italo-batista:

Sending productClick event to Pixel Manager with the purpose of usage in google-analytics.

#### Types of changes

* [X] New feature (a non-breaking change which adds functionality)

**Workspace**: https://persepolis--storecomponents.myvtex.com

### DIFF
--- a/react/components/GalleryItem.js
+++ b/react/components/GalleryItem.js
@@ -1,5 +1,6 @@
 import React from 'react'
 import { ExtensionPoint } from 'vtex.render-runtime'
+import { usePixel } from 'vtex.pixel-manager/PixelContext'
 import { path, sort, comparator, useWith, gt } from 'ramda'
 
 import { productShape } from '../constants/propTypes'
@@ -49,12 +50,19 @@ const GalleryItem = ({ item, displayMode, summary }) => {
     return normalizedProduct
   }
 
+  const { push } = usePixel()
+
+  const sendProductClickEvent = product => {
+    push({ event: 'productClick', product })
+  }
+
   return (
     <ExtensionPoint
       id="product-summary"
       {...summary}
       product={normalizeProductSummary(item)}
       displayMode={displayMode}
+      actionOnClick={() => sendProductClickEvent(item)}
     />
   )
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Added
+
+- Send productClick events to Pixel Manager.
+
 ## [3.17.8] - 2019-05-10
 ### Fixed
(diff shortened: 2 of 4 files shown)


react/components/GalleryItem.js
```diff
@@ -9,8 +10,8 @@ import { PropTypes } from 'prop-types'
 /**
  * Normalizes the item received in the props to adapt to the extension point prop.
  */
-const GalleryItem = ({ item, displayMode, summary }) => {
-  const normalizeProductSummary = product => {
+class GalleryItem extends Component {
```
lucasecdb: why did you move it to a class component?

react/components/GalleryItem.js
```diff
@@ -49,17 +50,29 @@ const GalleryItem = ({ item, displayMode, summary }) => {
...
+        product={this.normalizeProductSummary(item)}
+        displayMode={displayMode}
+        actionOnClick={() => this.sendProductClickEvent(item)}
+      />
+    )
+  }
 }
 
-GalleryItem.propTypes = {
+export const propTypes = {
```
lucasecdb: you removed the prop types from the component.

and why do you need to export this?

react/components/GalleryItem.js
italo-batista: Because the Pixel "interface" should have access to a component's props to push notifications.

react/components/GalleryItem.js
italo-batista: I removed because Pixel overwritten it anyway, but I exported to make it still available if other components need it.

react/components/GalleryItem.js
italo-batista: I will follow the suggestion of @jgfidelis and use usePixel. Didn't know its existance. :)

react/components/GalleryItem.js
lucasecdb: no, the pixel won't override the component's propTypes, it creates a _new_ component that wraps this one, the propTypes are not lost in the way

react/components/GalleryItem.js
lucasecdb: by the way, you can still use high order components in function components, its the same as you would use in a class. there was no need to convert it into a class even if the `usePixel` didn't exist

lucasecdb: check [this guide](https://reactjs.org/docs/higher-order-components.html) if you want to learn more

react/components/GalleryItem.js
```diff
@@ -49,12 +50,19 @@ const GalleryItem = ({ item, displayMode, summary }) => {
     return normalizedProduct
   }
 
+  const { push } = usePixel()
+
+  const sendProductClickEvent = product => {
+    push({ event: 'productClick', product: product })
```
jgfidelis:
```suggestion
    push({ event: 'productClick', product })
```

PR #175, opened 2019-05-17
Updated the context in the props

micas06gua:
The amount of items returned by search result is not accepting the configuration of the current documentation. Only by applying a context in the props of the component, the standard quantity of items returned by the search result works.

at least from the app I'm setting up, it worked this way.


jgfidelis:
Thank you for the help.

PR #176, opened 2019-05-17
use breadcrumb resolver from productSearch

jgfidelis:
https://fidelis--invictastores.myvtex.com/disney?map=ft
or
https://fidelis--invictastores.myvtex.com/collections?map=c%2CspecificationFilter_69

look at breadcrumb & at search title
#### Types of changes

### DIFF
--- a/react/components/SearchResult.js
+++ b/react/components/SearchResult.js
@@ -1,6 +1,7 @@
 import React, { Component } from 'react'
 import { Spinner } from 'vtex.styleguide'
 import { ExtensionPoint, withRuntimeContext } from 'vtex.render-runtime'
+import { compose, prop, last } from 'ramda'
 
 import LoadingOverlay from './LoadingOverlay'
 import { searchResultPropTypes } from '../constants/propTypes'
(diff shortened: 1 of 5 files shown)


react/components/SearchResult.js
```diff
@@ -152,6 +153,8 @@ class SearchResult extends Component {
     const hideFacets = !map || !map.length
     const showLoading = loading && !fetchMoreLoading
     const showContentLoader = showLoading && !showLoadingAsOverlay
+    const getLastName = compose(prop('name'), last)
```
klzns: Make this a method or an outside function so you don't recreate function on every render.

react/components/SearchResultContainer.js
```diff
@@ -132,9 +98,7 @@ const SearchResultContainer = props => {
         <ResultComponent
           {...props}
           showMore={showMore}
-          breadcrumbsProps={getBreadcrumbsProps(
-            Object.assign({}, params, { categoriesTrees, loading })
-          )}
+          breadcrumbsProps={{ categoryTree: breadcrumb }}
```
klzns: The name `categoryTree` still makes sense?

jgfidelis: Não quero causar breaking change ou ficar criando mais uma prop...

PR #177, opened 2019-05-21
Dispatch productImpression events to PixelManager

italo-batista:

As title says.
Also fix bug when pushing productClick event (param expected was not being sent)

https://italohmb--storecomponents.myvtex.com/

Click on a category gallery (ex: Notebooks). Events should be sent to Pixel.

#### Types of changes

* [X] New feature (a non-breaking change which adds functionality)

### DIFF
--- a/react/components/GalleryItem.js
+++ b/react/components/GalleryItem.js
@@ -56,13 +56,15 @@ const GalleryItem = ({ item, displayMode, summary }) => {
     push({ event: 'productClick', product })
   }
 
+  const product = normalizeProductSummary(item)
+
   return (
     <ExtensionPoint
       id="product-summary"
       {...summary}
-      product={normalizeProductSummary(item)}
+      product={product}
       displayMode={displayMode}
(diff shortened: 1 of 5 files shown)


react/components/GalleryItem.js
```diff
@@ -52,17 +52,33 @@ const GalleryItem = ({ item, displayMode, summary }) => {
 
   const { push } = usePixel()
 
-  const sendProductClickEvent = product => {
+  const pushPixelProductClickEvent = product => {
+    product.selectedSku = product.sku.itemId
```
jgfidelis: I don't  think it is safe to change it like this:

it would be safe to do:
```
const selectedProduct = {
  ...product,
  selectedSku: product.sku.itemId,
}
```

italo-batista: Thanks! I really appreciated this tip. But now I think the best place to do it is where the events will be received. So I deleted this line.

PR #178, opened 2019-05-22
add shop-review-interfaces to search results page

mvandeusen:
add support for new shop-review-interfaces app to place blocks on search results page. These blocks are expected to contain third party code from services that provide for reviews of merchants and/or their stores like Shopper Approved

added support for shop-review-interfaces blocks to interfaces and an extension point to place the blocks in the filters column on the search results page.

Allowing a review app which depends on shop-review-interfaces to place code onto the search results page

the Shopper Approved widget currently displays under the filters on this page: https://marc--sandboxusdev.myvtex.com/case?map=ft 

image

#### Types of changes

* [ x] New feature (a non-breaking change which adds functionality)


yarn.lock
```diff
@@ -0,0 +1,4 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
klzns: Remove this file

manifest.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "search-result",
   "vendor": "vtex",
-  "version": "3.18.0",
+  "version": "3.19.1",
```
klzns: Do not update the version just yet

klzns:
Changelog entry is missing.

store/blocks.json
```diff
@@ -8,5 +8,10 @@
       "total-products",
       "order-by"
     ]
+  },
+  "filter-navigator":{
```
jgfidelis: this is not correct, this would make `shop-review-widget` a default block for filter-navigator.

mvandeusen: What should I do to correct this so that the widget is allowed in the filter-navigator blocks? 

I've tried changing "blocks" to "allowed" and removing "filter-navigator" section from the blocks.json file completely. The former gives me an error stating that "allowed" is an unsupported field when I try to link the project and the latter results in my shop-review-widget not displaying inside the filter-navigator block in my workspace despite being defined as a block within my store-theme like this;

"search-result": {
    "blocks": [
      "filter-navigator",
      "gallery",
      "not-found",
      "breadcrumb",
      "total-products",
      "order-by"
    ]
  },
  "filter-navigator":{
    "blocks":[
      "shop-review-widget.shopper-approved-widget"
    ]
  },

polishq: Not sure why your store-theme isn't working, but you've defined the block as allowed in interfaces.json and that should be all you need to do. The blocks.json in this repo is basically just an "initial state" and would typically be overridden by the store-theme anyway.

klzns:
Fixed at #188

PR #192, opened 2019-06-17
Fix/cateogry css

jgfidelis:
Screen Shot 2019-06-17 at 10 37 43 AM
Screen Shot 2019-06-17 at 10 38 03 AM

### DIFF
--- a/react/components/DepartmentFilters.js
+++ b/react/components/DepartmentFilters.js
@@ -37,7 +37,7 @@ const DepartmentFilters = ({
       )}
       <div
         className={classNames(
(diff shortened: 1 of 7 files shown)


react/components/DepartmentFilters.js
```diff
@@ -37,7 +37,7 @@ const DepartmentFilters = ({
       )}
       <div
         className={classNames(
-          styles.selectedCategoriesContainer,
```
jgfidelis: this is not a breaking change because this class was never declared on the .css file, so this line was useless before

PR #194, opened 2019-06-18
create SearchQuery and use it in LocalQuery

jgfidelis:
Notar que:
https://fidelis--parquedpedro.myvtex.com/c/tng

Essa busca custom que utiliza a local query continua perfeitamente.


react/SearchQuery.js
```diff
@@ -0,0 +1,2 @@
+import SearchQuery from './components/SearchQuery'
```
kevinch: Missing changelog entry.

PR #210, opened 2019-07-26
fix price range navigation

jgfidelis:

By using the setQuery method, it skips some steps of the navigation method, maintains the page context and only triggers the refetch of the query with the new price range query param.

When setting a price range, it would redo the original query as a FT and not as a category, for example

https://fidteste--storecomponents.myvtex.com/apparel---accessories/d

#### Types of changes

### DIFF
--- a/react/components/PriceRange.js
+++ b/react/components/PriceRange.js
@@ -6,17 +6,14 @@ import { Slider } from 'vtex.styleguide'
 import { formatCurrency } from 'vtex.format-currency'
 
 import { facetOptionShape } from '../constants/propTypes'
-import {
-  getFilterTitle,
-  HEADER_SCROLL_OFFSET,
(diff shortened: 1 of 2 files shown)


react/components/PriceRange.js
```diff
@@ -35,18 +32,7 @@ const PriceRange = ({ title, facets, intl, priceRange }) => {
...
-
-      navigate({
-        to: window.location.pathname,
-        query: queryParams.toString(),
-        scrollOptions: {
-          baseElementId: 'search-result-anchor',
-          top: -HEADER_SCROLL_OFFSET,
-        },
-      })
+      setQuery({ priceRange: `${left} TO ${right}` }, { replace: true })
```
klzns:
```suggestion
      setQuery({ priceRange: `${left} TO ${right}` })
```

PR #212, opened 2019-07-30
Fix/search error

jgfidelis:
https://stime2--alssports.myvtex.com/

### DIFF
--- a/react/components/SearchQuery.js
+++ b/react/components/SearchQuery.js
@@ -86,13 +86,17 @@ const SearchQuery = ({
   }, [variables, page])
   return (
     <Query
+      key={variables.query}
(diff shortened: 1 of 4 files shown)


react/components/SearchQuery.js
```diff
@@ -86,13 +86,17 @@ const SearchQuery = ({
   }, [variables, page])
   return (
     <Query
+      key={variables.query}
```
lbebber:
```suggestion
      /** This key fixes an issue where data from the previous query
       * would persist after changing `variables`, while loading.
       * https://github.com/apollographql/react-apollo/issues/2202 */
      key={variables.query}
```

PR #246, opened 2019-09-26
allow-custom-options-in-order-by

hiagolcm:

The sort options in `OrderBy` are hard coded. This PR aims to offer the `customOptions` prop to allow custom options.

It is also related to https://github.com/vtex-apps/store-discussion/issues/101

[workspace](https://customorderbyoptions--biggy.myvtex.com/)

In this example only three options are available.

```
"order-by.v2#custom": {
    "props": {
      "customOptions": [
        {
          "value": "",
          "label": "store/ordenation.release.date"
        },
        {
          "value": "OrderByTopSaleDESC",
          "label": "store/ordenation.sales"
        },
        {
          "value": "OrderByReleaseDateDESC",
          "label": "store/ordenation.release.date"
        }
      ]
    }
  }
```
image

#### Types of changes

### DIFF
--- a/react/OrderByFlexible.js
+++ b/react/OrderByFlexible.js
@@ -5,7 +5,7 @@ import OrderBy from './OrderBy'
 
 import styles from './searchResult.css'
 
-const withSearchPageContextProps = Component => () => {
+const withSearchPageContextProps = Component => (props) => {
   const { orderBy } = useSearchPage()
   if (orderBy == null) {
     return null
@@ -13,7 +13,7 @@ const withSearchPageContextProps = Component => () => {
 
   return (
     <div className={styles['orderBy--layout']}>
-      <Component orderBy={orderBy} />
+      <Component {...props} orderBy={orderBy} />
     </div>
   )
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 ### Added
-- `customOptions` to `OrderBy`
+- `hiddenOptions` to `OrderBy`
 
 ### Changed
 - The title in `SelectionListOrderBy` now returns `''` instead of an error when the value is not found.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 ### Added
-- `hiddenOptions` to `OrderBy`
+- `hiddenOptions` prop to `OrderBy`.
 
 ### Changed
 - The title in `SelectionListOrderBy` now returns `''` instead of an error when the value is not found.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- `customOptions` to `OrderBy`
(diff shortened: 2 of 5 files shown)


victorhmp:
Hey @hiagolcm!

So glad to see this PR since I was about to start implementing this! I have a few suggestions regarding the API for this feature that I think would make it a bit more user-friendly (and maybe even reduce the change of something going wrong).

I would not expect the user to write out the options that they want to show exactly as it is in the code, as this is a bit error-prone. Since we don't support any custom `orderBy` value, I would expect the user to write an array of values that they want to enable, and that's it. No need to pass in the `label` for example and the values you would expect to be inside the `customOptions` array are an `enum`, right? Much more predictable! It would go something like:

```json
"order-by.v2#custom": {
    "props": {
      "customOptions": [ 
          "OrderByReleaseDateDESC", "OrderByTopSaleDESC", "OrderByReleaseDateDESC"
      ]
    }
  }
```

hiagolcm:
> Hey @hiagolcm!
> 
> So glad to see this PR since I was about to start implementing this! I have a few suggestions regarding the API for this feature that I think would make it a bit more user-friendly (and maybe even reduce the change of something going wrong).
> 
> I would not expect the user to write out the options that they want to show exactly as it is in the code, as this is a bit error-prone. Since we don't support any custom `orderBy` value, I would expect the user to write an array of values that they want to enable, and that's it. No need to pass in the `label` for example and the values you would expect to be inside the `customOptions` array are an `enum`, right? Much more predictable! It would go something like:
> 
> ```json
> "order-by.v2#custom": {
>     "props": {
>       "customOptions": [ 
>           "OrderByReleaseDateDESC", "OrderByTopSaleDESC", "OrderByReleaseDateDESC"
>       ]
>     }
>   }
> ```

@victorhmp

I'm glad to be helpful and thank you for your suggestions.

The main reason of this PR is to provide a way to integrate new search engines. In the current situation, VTEX-IO components only supports the default VTEX search.
Here is another PR I submitted with the same goal https://github.com/vtex-apps/store-components/pull/589

This PR aims to make the sort options customizable. By customizable I mean: provide new options (from a new search engine) or hide VTEX default options. In this way, if only the value is provided in the props, as you suggested, new search engines will be limited to the VTEX options.

Although your suggestion doesn't cover all the cases I initially proposed, I will keep it. Configurations with so many details of the implementation should not be written in the blocks.

In a conversation with the store-components team, we decided to reduce the scope of this PR and only provide a way to hide VTEX options (not create new ones anymore). Thus, I changed the prop from `customOptions` to `hiddenOptions`.

react/OrderByFlexible.js
```diff
@@ -5,15 +5,15 @@ import OrderBy from './OrderBy'
...
+const withSearchPageContextProps = Component => (props) => {
   const { orderBy } = useSearchPage()
   if (orderBy == null) {
     return null
   }
 
   return (
     <div className={styles['orderBy--layout']}>
-      <Component orderBy={orderBy} />
+      <Component {...props} orderBy={orderBy} />
```
lbebber: what is the purpose of this change?

lbebber: by that I mean, can you just pass down the specific prop you need?

hiagolcm: You mean this?

```
const withSearchPageContextProps = Component => ({hiddenOptions}) => {
 // omitted
<Component hiddenOptions={hiddenOptions} orderBy={orderBy} />
// omitted
}
```

README.md
```diff
@@ -397,6 +397,25 @@ Notice that the default behavior for your store will be the one defined by the `
 
 Also, you can configure the product summary that is defined on search-result. See [here](https://github.com/vtex-apps/product-summary/blob/master/README.md#configuration) the Product Summary API.
 
+##### `order-by` block
```
iaronaraujo: I think you can remove this file since we moved the readme to the docs folder

PR #247, opened 2019-10-02
Handle errors in the fetchMore promise

iaronaraujo:

Sometimes when scrolling through search results an error would happen, making the page load indefinitely or by showing the user a "products not found" page. These errors were caused by the `fetchMore` promise that was called on some `useFetchMore`'s methods that are used on the fetch more buttons and infinite scrolling spinner. This PR handles possible exceptions thrown by this promise, making these errors almost unnoticeable by the users.

The error is very hard to reproduce since it happens randomly when you open a new search result and try browsing through the results, but here is the [workspace](https://iaronaraujo--alssports.myvtex.com/clothing/pants?map=c,c&order=OrderByPriceASC&page=8)

#### Types of changes

### DIFF
--- a/react/components/SearchResult.js
+++ b/react/components/SearchResult.js
@@ -131,6 +131,7 @@ class SearchResult extends Component {
       mobileLayout,
       isMobile,
       pagination,
+      infiniteScrollError,
     } = this.props
     const {
       mobileLayoutMode,
@@ -234,7 +235,7 @@ class SearchResult extends Component {
                 <ExtensionPoint id="not-found" />
               </div>
             )}
-            {pagination === PAGINATION_TYPE.SHOW_MORE ? (
+            {pagination === PAGINATION_TYPE.SHOW_MORE || infiniteScrollError ? (
               <FetchMoreButton
                 products={products}
                 to={to}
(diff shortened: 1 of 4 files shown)


react/components/SearchResult.js
```diff
@@ -234,7 +235,7 @@ class SearchResult extends Component {
                 <ExtensionPoint id="not-found" />
               </div>
             )}
-            {pagination === PAGINATION_TYPE.SHOW_MORE ? (
+            {pagination === PAGINATION_TYPE.SHOW_MORE || infiniteScrollError ? (
```
victorhmp: Can this error go away with a retry?

iaronaraujo: Yes! The reason for this line is because when there is an error, you can't trigger the infinite scrolling again, so it gets impossible to call `fetchMoreNext` once an error appears with infinite loading. This line makes sure a button appears on these cases and then you go back to infinite scrolling once you have fetched more results successfully since this resets the infinite scrolling behavior!

react/hooks/useFetchMore.js
```diff
@@ -39,6 +40,11 @@ const handleFetchMore = (
       setLoading(false)
       fetchMoreLocked.current = false
 
+      if (!prevResult || !fetchMoreResult) {
+        updateQueryError.current = true
```
lbebber: a function ideally shouldn't modify an external object.

this works, but can be confusing. should at least be marked as a TODO, and fixed soon

PR #253, opened 2019-10-10
Trigger reload on price-range change and reset pagination

victorhmp:

Fix undesired behavior regarding updates to the `price-range` in a `search-result` page. 

- Changes in price-range in the current search would not trigger a reload and would not provide any feedback to the user, resulting in bad UX.
- If a change in price-range was made on a certain search-result page, the pagination would not reset.

Change the `price-range` and see that a reload is triggered and a new search is made.

https://loadingpricerange--storecomponents.myvtex.com/apparel---accessories

#### Types of changes

### DIFF
--- a/react/components/PriceRange.js
+++ b/react/components/PriceRange.js
@@ -13,7 +13,7 @@ const DEBOUNCE_TIME = 500 // ms
 
 /** Price range slider component */
 const PriceRange = ({ title, facets, intl, priceRange }) => {
-  const { culture, setQuery } = useRuntime()
+  const { culture, navigate } = useRuntime()
 
   const navigateTimeoutId = useRef()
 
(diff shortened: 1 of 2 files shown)


react/components/PriceRange.js
```diff
@@ -23,7 +23,14 @@ const PriceRange = ({ title, facets, intl, priceRange }) => {
     }
 
     navigateTimeoutId.current = setTimeout(() => {
-      setQuery({ priceRange: `${left} TO ${right}`, page: undefined })
+      const urlParams = new URLSearchParams(window.location.search)
```
klzns: Have you tested in Safari and checked the support of this? Or if we poly fill?

victorhmp: Seems to work fine on iOS

victorhmp: (Talked to Breno in-person, this is just to keep a record) The problem with using `URLSearchParams` is regarding SSR, which would not be a problem here

PR #260, opened 2019-10-29
Refactored `useFetchMore`

iaronaraujo:

Previously every page related data had an `useState`. This PR refactors this logic to use an `useReducer` instead!

[Workspace](https://iaronaraujo--alssports.myvtex.com/clothing/pants?map=c,c&order=OrderByNameASC)

#### Types of changes

### DIFF
--- a/react/hooks/useFetchMore.js
+++ b/react/hooks/useFetchMore.js
@@ -1,4 +1,4 @@
-import { useState, useRef, useCallback, useEffect } from 'react'
+import { useState, useRef, useCallback, useEffect, useReducer } from 'react'
 import { min, max } from 'ramda'
 import { useRuntime } from 'vtex.render-runtime'
 import {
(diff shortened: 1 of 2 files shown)


react/hooks/useFetchMore.js
```diff
@@ -11,6 +11,39 @@ export const FETCH_TYPE = {
...
+      }
+    case 'PREVIOUS_PAGE':
+      return {
+        ...state,
+        page: state.previousPage,
+        previousPage: state.previousPage - 1,
+        from,
+      }
+
+    case 'ROLLBACK':
```
jgfidelis: this looks wrong, it is not changing anything, you should pass the variables to rollback to in here

react/hooks/useFetchMore.js
```diff
@@ -157,23 +187,19 @@ export const useFetchMore = props => {
     )
     //if error, rollback
     if (promiseResult && updateQueryError.current) {
-      setCurrentTo(currentTo)
-      setCurrentPage(currentPage)
-      setNextPage(nextPage)
-      setQuery({ page: currentPage }, { replace: true })
+      pageDispatch({ type: 'ROLLBACK' })
```
jgfidelis: this is weird, did you test it?

iaronaraujo: Yes! This works the same way as `useState`, meaning the values only actually change after the render, so since the rollback is done before the render, this just restores the state to be as it used to be!

jgfidelis: hmmm. I think this is trusting too much in implementation detail. If I would do it myself, I would pass the args in to rollback in this event as well, to be more declarative. I leave this to the team to decide what is better for their codebase.

PR #262, opened 2019-10-30
Use custom distribution any time

grupo-exito-ecommerce:
Variable creation to indicate if the default vtex distribution is used or if the distribution sent by the maxItemsPerRow field is always used

Always be able to use the distribution indicated in the product gallery

It is necessary that the distribution shown in the search-result is a defined one and not calculated according to the current screen

Send a property called useDefaultDistribution to always use the value of maxItemsPerRow

``` javascript
 return cloneElement(children, {
    maxItemsPerRow: dynamicDistribution,
    useDefaultDistribution: false
  });
```

`useDefaultDistribution` by default is true

#### Types of changes

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -32,7 +32,8 @@
     "vtex.flex-layout": "0.x",
     "vtex.tab-layout": "0.x",
     "vtex.search-page-context": "0.x",
-    "vtex.blog-interfaces": "0.x"
+    "vtex.blog-interfaces": "0.x",
+    "vtex.responsive-values": "0.x"
   },
   "$schema": "https://raw.githubusercontent.com/vtex/node-vtex-api/master/gen/manifest.schema"
 }
--- a/manifest.json
+++ b/manifest.json
@@ -5,9 +5,7 @@
   "title": "VTEX Search Result",
   "description": "A search result wrapper component",
   "mustUpdateAt": "2019-04-25",
-  "registries": [
-    "smartcheckout"
-  ],
+  "registries": ["smartcheckout"],
   "scripts": {
     "postreleasy": "vtex publish --verbose"
   },
--- a/react/Gallery.js
+++ b/react/Gallery.js
@@ -28,6 +28,7 @@ const Gallery = ({
   width,
   summary,
   showingFacets,
+  useDefaultDistribution = true
 }) => {
   const { isMobile } = useDevice()
 
@@ -38,7 +39,7 @@ const Gallery = ({
     return maxItemsPerRow <= maxItems ? maxItemsPerRow : maxItems
   }
 
-  const itemsPerRow =
+  const itemsPerRow = !useDefaultDistribution ? maxItemsPerRow :
     layoutMode === 'small'
       ? TWO_COLUMN_ITEMS
       : isMobile
@@ -90,6 +91,7 @@ Gallery.propTypes = {
   /** Min Item Width. */
   minItemWidth: PropTypes.number,
   showingFacets: PropTypes.bool,
+  useDefaultDistribution: PropTypes.bool
 }
 
 export default withResizeDetector(Gallery)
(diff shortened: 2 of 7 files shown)


react/Gallery.js
```diff
@@ -38,7 +39,7 @@ const Gallery = ({
     return maxItemsPerRow <= maxItems ? maxItemsPerRow : maxItems
   }
 
-  const itemsPerRow =
+  const itemsPerRow = !useDefaultDistribution ? maxItemsPerRow :
```
victorhmp: There's a side-effect to this implementation.
I don't know if you want this behavior, but if a user were to set this `useDefaultDistribution` prop to `false`, the `LayoutSwitcher` component would be useless, as this conditional check would not try to validate `layoutMode === 'small'`.

victorhmp: Also, I think this bit of code looks confusing with 3 ternary conditions in a row, a refactor would be welcome.

klzns:
With that solution, the mobile layout might get really packed:

Mobile:
image

How would we deal with that scenario?

grupo-exito-ecommerce:
> Con esa solución, el diseño móvil podría estar realmente lleno
> 
> ¿Cómo lidiaríamos con ese escenario?
> 
> Móvil:
> imagen


So that the distribution does not look bad on mobile, we are using a `currentWidth` variable that tells us what the current resolution is and we have an `productDistribution` array that has the settings for each resolution

``` javascript
const currentWidth = useWidth();

return cloneElement(children[0], {
    maxItemsPerRow: productDistribution[currentWidth],
    useDefaultDistribution: false
  });
```

`useWidth` is a hook that returns the current resolution` `xs' | 'sm' | 'md' | 'lg' | 'xl'`
and `productDistribution` is an array that contains `[xs: 2, sm: 3, md: 4, lg: 5, xl: 6]`

manifest.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "search-result",
   "vendor": "vtex",
-  "version": "3.35.7",
+  "version": "3.35.8",
```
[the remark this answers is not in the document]
grupo-exito-ecommerce: Is necesary update the CHANGELOG.MD file or not?

react/Gallery.js
grupo-exito-ecommerce: Check this change.

``` javascript
  const defaultDistribution = layoutMode === 'small'
  ? TWO_COLUMN_ITEMS
  : isMobile
  ? ONE_COLUMN_ITEMS
  : getItemsPerRow() || maxItemsPerRow

  const itemsPerRow = !useDefaultDistribution ? maxItemsPerRow : defaultDistribution
```

klzns:
Oh, I see.

In that case I think we should change `maxItemsPerRow` to be a [responsive-value](https://github.com/vtex-apps/responsive-values)

grupo-exito-ecommerce:
> Oh, I see.
> 
> In that case I think we should change `maxItemsPerRow` to be a [responsive-value](https://github.com/vtex-apps/responsive-values)
> 
> That way you wouldn't need to use that hook

Okay, we will use this configuration from our implementation, but we need to be able to always use our configuration. They can validate if the change made is fine to be able to approve the integration.

manifest.json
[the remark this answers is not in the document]
victorhmp: Yep, it is

react/Gallery.js
victorhmp: Nice! That solves the layout on mobile :)
Would you guys please just run `yarn lint --fix`?

grupo-exito-ecommerce: Sorry, We run `yarn  lint --fix`  in the project but get this error

 ```
yarn run v1.19.1
error Command "lint" not found.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

The search-result project only have this scrips

```
"scripts": {
    "pretest": "yarn",
    "test": "vtex-test-tools test"
  },
```

We run the `test` script and no have errors with the change. 
Is what was done right or wrong? or what should I do

klzns:
@allcontributors add @grupo-exito-ecommerce for code

klzns:
@allcontributors add @grupo-exito-ecommerce for code

PR #272, opened 2019-11-11
Fix prevent route change

iaronaraujo:

Fixed some issues related to `preventRouteChange`

[Als](https://mytest2--alssports.myvtex.com/clothing/Womens?map=c,specificationFilter_7721)
[Boticario](https://iaronaraujo--boticario.myvtex.com/perfumaria/10%20ml/365%20ml?map=c,specificationFilter_120,specificationFilter_120)

#### Types of changes

### DIFF
--- a/react/components/CategoryFilter.js
+++ b/react/components/CategoryFilter.js
@@ -10,10 +10,6 @@ import CategoryItem from './CategoryItem'
 
 import styles from '../searchResult.css'
 
-/* Since Category isn't really on the same abstraction level as the other 
-facets, it was decided that it should not have support for preventRouteChange */
-const PREVENT_ROUTE_CHANGE = false
-
 const getSelectedCategories = rootCategory => {
   let node = rootCategory.children
   const selectedCategories = [rootCategory]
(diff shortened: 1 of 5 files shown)


react/components/CategoryFilter.js
```diff
@@ -28,13 +28,9 @@ const getSelectedCategories = rootCategory => {
...
 
-const CategoryFilter = ({
-  category,
-  shallow = false,
-  onCategorySelect,
-  preventRouteChange = false,
-}) => {
+const CategoryFilter = ({ category, shallow = false, onCategorySelect }) => {
   const { map } = useContext(QueryContext)
+  const preventRouteChange = false
```
klzns: Justify that with a comment

klzns: Why is that different from DepartmentFilter?

jgfidelis: why define this variable here, and not pass `false` directly?

jgfidelis: or define it outside of the function

iaronaraujo: @klzns @jgfidelis  when talking to Bebber we decided that categories shouldn't support `preventRouteChange` since it doesn't really work as a facet. I can define it outside the function as a constant though.

klzns: Why department doesnt need this code?

klzns: What happens if you just dont pass the second param to the function?

PR #287, opened 2019-12-22
Blocks mobile validation of filter-navigator

gustavopvasconcellos:
Makes `filter-navigator` more customizable on mobile view if new possibilities to use with `drawer`, `flex-layout` and all others interfaces of search-result.

On layout's mobile view have the same desktop functionallity and layout but inside of a ´drawer´. The `isMobile` validation makes impossible to keep the same behavior and results in an unexpected layout.

There is a workspace without CSS rules: https://alwaysondesktop--unimarcdev.myvtex.com/search?_query=leche
Its just test on a mobile device to check the same behavior of desktop mode.

Layout:
image

With the new props:
image

#### Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+- Add a new props to `filter-navigator.v3` blocks the mobile validation
 
 ## [3.39.10] - 2019-12-23
 ### Changed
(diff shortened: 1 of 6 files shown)


klzns:
Can you provide an image of how you would like it to be after customisations?

klzns:
Also, thanks for the pull requests! They are welcome!

gustavopvasconcellos:
@klzns Of course, its should be like this:

Drawer as a simple button:
image

Drawer content with `flex-layout` and inside it: `search-title.v2`, `total-products.v2` and `filter-navigator` with the new prop:
image

gustavopvasconcellos:
@klzns By default, this new prop is `false`, so the same behavior of `filter-navigator` stills working.

PR #299, opened 2020-01-17
New search urls

Jeymisson:
What problem is this solving?
Resolve more friendly URLS for filters

In order to have a platform more SEO friendly we are changing our searches URL structure when applying filters. This is a workaround that should be resolved when we have a complete solution with collections where each collection will be a filter with a nice structured URL.

**Meanwhile...**

How should this be manually tested?
Open category pages and navigate. alssporst has hardcoded map=c,c,c values on menu.
We are not normalizing this at the menu component.
https://jey--alssports.myvtex.com

#### Types of changes

### DIFF
--- a/react/hooks/useFacetNavigation.js
+++ b/react/hooks/useFacetNavigation.js
@@ -32,20 +32,9 @@ const removeSelectedFacets = selectedFacet => {
   }
 }
 
-const isLegacySearchFormat = (query, mapSegments, pathSegments) => {
-  return (
-    query.includes(SPEC_FILTER) ||
-    (mapSegments && mapSegments.length === pathSegments.length)
-  )
-}
-
 const removeElementAtIndex = (str, index) => str.filter((_, i) => i !== index)
 
-const newFacetPathName = (facet, isLegacySearch) => {
-  if (isLegacySearch && facet.selected) {
-    return facet.value
-  }
-
+const newFacetPathName = facet => {
   return facet.map && facet.map.includes(SPEC_FILTER)
     ? `${facet.title
         .replace(/\s/g, SPACE_REPLACER)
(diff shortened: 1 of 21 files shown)


react/hooks/useFacetNavigation.js
```diff
@@ -5,109 +5,177 @@ import { useQuery } from '../components/QueryContext'
...
+
+    navigate({
+      to: `${PATH_SEPARATOR}${currentQuery}`,
+      query: urlParams.toString(),
+      scrollOptions,
+      modifiersIgnore,
+    })
+  })
+
+  navigateToFacet.storeSelectedFacets = storeSelectedFacets
```
jgfidelis: I get what you did but modifying the object of a `useCallback` call is completely against React pattern. You can make this hook return a tuple [navigateToFacet, storeSelectedFacets], something like that, but do change navigateToFacet like this

react/hooks/useSelectedFilters.js
```diff
@@ -1,15 +1,12 @@
...
 
 /**
  * This hook is required because we make the facets query
  * with only the categories and fulltext parameters, so we
  * need to calculate manually if the other filters are selected
  */
-const useSelectedFilters = facets => {
-  const { query, map } = useContext(QueryContext)
+const useSelectedFilters = (facets, queryArgs) => {
+  const { query, map } = queryArgs
```
jgfidelis: the usage of a context is because this data is used everywhere and in a lot of places and so we don't  want to pass this everywhere. Why the need for queryArgs, can't you set the appropriate value in context? That would be the correct

jgfidelis:
Você precisa testar:
Desktop
Mobile
Ter certeza que toda parte de ordenação e de página da search continua funcionando

PR #322, opened 2020-03-11
Fix mobile facets

iaronaraujo:

Previously, on mobile, clicking on a specification filter facet would break the sidebar. With this PR this is fixed, but the facet navigation behavior is still weird

[Workspace](https://iaronaraujo--eriksbikeshop.myvtex.com/cycling/bicycles/road/Dropbar/Performance?map=c,c,c,specificationFilter_254,specificationFilter_254)

1. open this on mobile
2. click on filter > color family > black. Notice that the interface won't show that it was checked.
3. click on "apply"
4. notice that the filter worked as intended.

#### Types of changes

### DIFF
--- a/react/components/AccordionFilterGroup.js
+++ b/react/components/AccordionFilterGroup.js
@@ -1,5 +1,5 @@
 import React from 'react'
-import { injectIntl } from 'react-intl'
+import { useIntl } from 'react-intl'
 
 import AccordionFilterItem from './AccordionFilterItem'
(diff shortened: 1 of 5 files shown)


react/components/AccordionFilterGroup.js
```diff
@@ -1,9 +1,12 @@
 import React from 'react'
+import { injectIntl } from 'react-intl'
```
klzns: Use useIntl instead

PR #328, opened 2020-03-19
adjust compatibility with the new search protocol

hiagolcm:

Currently, VTEX has its own protocol between the search API and the UX components. This protocol forces new search-engines to implement specific rules to communicate with the UI. The search-protocol project purposes a generic way to make the UI/API communication.

This PR adds a "compatibility layer" to handle the new `search-protocol`. With this layer, there is no need to make further changes in the UI components.

It also adds three new props that are essential for new search engines. They are: `fuzzy`, `operator` and `searchState`. These props are in the URL as queryStrings.

Finally, this PR makes a slight improvement in the FilterNavigator by avoiding unnecessary mounting.

I have created two workspaces, each one with a search engine
- [`search-graphql` engine](https://hiago--storecomponents.myvtex.com/)
- [`vtex-search` engine](https://hiagolucas--storecomponents.myvtex.com/) - It only solves full text searches.

#### Types of changes

#### Notes
This PR depends on 
- https://github.com/vtex-apps/search-graphql/pull/64
- https://github.com/vtex-apps/store-resources/pull/107
- https://github.com/vtex-apps/store/pull/440
Don't publish it before them!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+## [3.53.2] - 2020-04-03
 ### Fixed
 - Added missing blocks on README and fixed some nitpicks.
 
--- a/manifest.json
+++ b/manifest.json
@@ -1,7 +1,7 @@
 {
   "name": "search-result",
   "vendor": "vtex",
-  "version": "3.53.1",
+  "version": "3.53.2",
   "title": "VTEX Search Result",
   "description": "A search result wrapper component",
   "mustUpdateAt": "2019-04-25",
--- a/react/package.json
+++ b/react/package.json
@@ -48,5 +48,5 @@
     "regenerator-runtime": "^0.13.1",
     "typescript": "3.8.3"
   },
-  "version": "3.53.1"
+  "version": "3.53.2"
 }


react/components/SearchQuery.js
```diff
@@ -50,22 +57,52 @@ const useCombinedRefetch = (productRefetch, facetsRefetch) => {
   )
 }
 
-const useCorrectPage = ({ page, query, map, orderBy }) => {
-  const pageRef = useRef(page)
+const isCurrentDifferent = (ref, currentVal) => ref.current !== currentVal
+
+const useShouldResetPage = (query, map, orderBy) => {
   const queryRef = useRef(query)
```
jgfidelis: I think this might be a moment to refactor all of these refs to use a reducer state model.

Then you can just dispatch the reset action when needed. It may be done in a later PR though.

react/hooks/useFacetNavigation.js
```diff
@@ -167,11 +169,16 @@ const useFacetNavigation = selectedFacets => {
...
 
       if (preventRouteChange) {
-        setQuery({
+        const queries = {
           map: `${currentMap}`,
           query: `/${currentQuery}`,
           page: undefined,
-        })
+        }
+        if (fuzzy) queries.fuzzy = fuzzy
```
jgfidelis: better to keep pattern and use `if (x) { fn() }`

react/hooks/useFetchMore.js
```diff
@@ -199,7 +216,13 @@ export const useFetchMore = props => {
     const from = max(0, to - maxItemsPerPage + 1)
     setInfiniteScrollError(false)
     pageDispatch({ type: 'PREVIOUS_PAGE', args: { from } })
-    setQuery({ page: pageState.previousPage }, { replace: true, merge: true })
+
+    const queries = { page: pageState.previousPage }
+    if (fuzzy) queries.fuzzy = fuzzy
```
jgfidelis: you should abstract this logic in a function, its used like 5 times.

hiagolcm: I've just realized that I can accomplish the same result by using `fuzzy: fuzzy || fuzzy`.
I think it is better now.

react/FilterNavigator.js
```diff
@@ -96,82 +96,81 @@ const FilterNavigator = ({
 
   const filterClasses = classNames({
     'flex items-center justify-center flex-auto h-100': mobileLayout,
+    dn: loading,
   })
 
-  if (loading && !mobileLayout) {
```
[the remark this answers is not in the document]
hiagolcm: Now, this statement is inside a `Fragment` and I can't use the `if` inside the curly braces.
Do you think I should create a new function to handle this JSX? This way I would be able to use the `if` again.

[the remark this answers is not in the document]
hiagolcm: Now, this statement is inside a `Fragment`, and I can't use the `if` inside the curly braces.
Do you think I should create a new function to handle this JSX? This way, I would be able to use the `if` again.

react/components/SearchQuery.js
hiagolcm: Nice suggestion!
I have just created an issue about that.
https://github.com/vtex-apps/search-result/issues/330

I could make this PR by myself when I have a chance.

hiagolcm:
> Is this linked in this [workspace](https://hiagolucas--storecomponents.myvtex.com/apparel---accessories/)? If so, non-ft pages seem to be broken. Also, have you rebased this app? There may be some conflicts with the new URL change.

@iaronaraujo I forgot to mention that `vtex-search` can't solve non-ft pages yet 🤦‍♂ . So broken non-ft pages are totally expected. I only added this workspace to show that now it is possible to set a different search engine.

I have rebased the app today, the last release was the v3.51.4. It looks ok to me!

PR #333, opened 2020-03-26
Feat/installment

jgfidelis:

Add installmentCriteria prop to search query.
Make it editable in site editor
add docs

#### Types of changes


react/components/SearchQuery.js
```diff
@@ -6,8 +6,13 @@ import searchMetadataQuery from 'vtex.store-resources/QuerySearchMetadata'
 import facetsQuery from 'vtex.store-resources/QueryFacets'
 
 const DEFAULT_PAGE = 1
-const DEFAULT_SKU_FILTER = 'ALL_AVAILABLE'
-const DEFAULT_FACETS_BEHAVIOR = 'Static'
+
+const DEFAULT_QUERY_VALUES = {
```
lbebber: (nitpick) I don't think there's gonna be any actual problems with this, but the values of this object are mutable, so treating this as a constant might be a bit misleading.

I'd just create different constants for each of its items, instead of grouping them into an object

(this is just a suggestion)

docs/README.md
```diff
@@ -380,6 +380,7 @@ These properties can be changed in the `blocks.json` file of your theme.
...
 | `skusFilter`           | `SkusFilterEnum` | Control SKUs returned for each product in the query. The less SKUs needed to be returned, the more performant your shelf query will be.                                                               | `"ALL_AVAILABLE"` |
 | `simulationBehavior`           | `SimulationBehaviorEnum` | Set to "skip" value if you want faster queries and do not care about most up to date prices or stock.                                                               | `"default"` |
+| `installmentCriteria`               | `InstallmentCriteriaEnum`                 | ControlControl what price to be shown when price has different installments options.                                                                                                    | `"MAX_WITHOUT_INTEREST"` |
```
lbebber:
```suggestion
| `installmentCriteria`               | `InstallmentCriteriaEnum`                 | Control what price to be shown when price has different installments options.                                                                                                    | `"MAX_WITHOUT_INTEREST"` |
```

PR #360, opened 2020-05-21
add hidden field

hiagolcm:

There are some filters that should be hidden in UI. For example, the ones related to collections. This PR adds a `hidden` field to indicate which filter should be hidden.

⚠️ Please, do not merge it before:
- https://github.com/vtex-apps/search-resolver/pull/29
- https://github.com/vtex-apps/search-resolver/pull/28
- https://github.com/vtex-apps/search-graphql/pull/82
- https://github.com/vtex-apps/store-resources/pull/119

[Workspace](https://hiago--storecomponents.myvtex.com/apparel---accessories/clothing/)

#### Checklist/Reminders

- [ ] Updated `README.md`.
- [x] Updated `CHANGELOG.md`.
- [ ] Linked this PR to a Clubhouse story (if applicable).
- [ ] Updated/created tests (important for bug fixes).
- [ ] Deleted the workspace after merging this PR (if applicable).

#### Type of changes

✔️ | Type of Change
---|---
_ | Bug fix 
✔️ | New feature 
_ | Breaking change 
_ | Technical improvements

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Hide facets with property `hidden`.
 
 ## [3.59.3] - 2020-06-01
 ### Added
--- a/react/FilterNavigator.js
+++ b/react/FilterNavigator.js
@@ -82,7 +82,10 @@ const FilterNavigator = ({
     const options = [
       ...specificationFilters.map(filter => {
         return filter.facets.map(facet => {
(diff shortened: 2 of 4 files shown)


hiagolcm:
> Missing changelog!
> Also, what's the difference between this and the hidden facet prop? https://github.com/vtex-apps/search-result#available-props-for-search-result-layoutdesktop-search-result-layoutmobile-and-search-not-found-layout

Oh, sorry about the changelog 🤦 

The difference is that this new hidden is set in the backend. We have some rules about which facets to show. For example, the `productClusterIds` should not be shown to the user, but it needs to be present in the URL. So we just send it to the frontend with `hidden: true`.

If we use the `hiddenProps`, we would need to hardcode these rules in every store-theme.

Summarizing everything, `hiddenProps` will be used when the client wants to customize a specific filter, while this new `hidden` prop will be used to apply backend rules.

PR #384, opened 2020-06-30
Missing filter items when there are no options selected.

hiagolcm:

When the user unselects all facets on mobile, the options disappear.

GgaTYd4kg2

#### How to test it?

[Workspace](https://facetdisappear--storecomponents.myvtex.com/apparel---accessories/)

### DIFF
--- a/react/hooks/useSelectedFilters.js
+++ b/react/hooks/useSelectedFilters.js
@@ -1,4 +1,4 @@
-import { zip } from 'ramda'
+import { zip, isNil } from 'ramda'
 import { useFilterNavigator } from '../components/FilterNavigatorContext'
 import { isSameMap } from '../utils/queryAndMapUtils'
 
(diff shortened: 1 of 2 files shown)


react/hooks/useSelectedFilters.js
```diff
@@ -9,7 +9,7 @@ import { isSameMap } from '../utils/queryAndMapUtils'
  */
 const useSelectedFilters = facets => {
   const { query, map } = useFilterNavigator()
-  if (!query && !map) {
+  if (isNil(query) && isNil(!map)) {
```
Klynger, line 12: `isNil(!map)` is always `false`

kaisermann: We are also discouraging the overall use of `ramda` in our react apps (check the newest [eslint config here](https://github.com/vtex/typescript/blob/master/packages/eslint-config-vtex-react/io.js#L21)).

PR #385, opened 2020-06-30
add facets breadcrumb as fallback

hiagolcm:

The breadcrumb has two different origins: the `productSearch` query, and the `facets` query. Each search-engine decides how to implement it.
The problem is that the `SearchResult` is not considering the `facets` breadcrumb as a fallback.

#### How to test it?

[Workspace](https://hiago--redoxx.myvtex.com/bags/duffel?map=c,specificationFilter_818&page=1)

Before
image

After
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+- `SearchResult` doesn't use the `facets` breadcrumb as fallback.
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+- `SearchResult` doesn't use the `facets` breadcrumb as fallback.
```
klzns, line 11:
```suggestion
- `SearchResult` use the `facets` breadcrumb as fallback.
```

PR #389, opened 2020-07-13
New CSS handles on `order-by` selected item.

BeatrizMiranda:
Be able to change the style of the selected order by item. 

#### How to test it?

[Workspace](https://category--menudigital.myvtex.com/Vista-Corona?order=OrderByScoreDESC)

image

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -392,7 +392,7 @@ In order to apply CSS customization in this and other blocks, follow the instruc
 | `orderByButton`                       |
 | `orderByDropdown`                     |
 | `orderByOptionItem`                   |
-| `orderByOptionItem--{selectedItem}`         |
+| `orderByOptionItem--selected`         |
 | `orderByOptionsContainer`             |
 | `orderByText`                         |
 | `orderBy`                             |
--- a/docs/README.md
+++ b/docs/README.md
@@ -392,7 +392,7 @@ In order to apply CSS customization in this and other blocks, follow the instruc
 | `orderByButton`                       |
 | `orderByDropdown`                     |
 | `orderByOptionItem`                   |
-| `orderByOptionItem--selected`         |
+| `orderByOptionSelected`               |
 | `orderByOptionsContainer`             |
 | `orderByText`                         |
 | `orderBy`                             |
--- a/docs/README.md
+++ b/docs/README.md
@@ -392,6 +392,7 @@ In order to apply CSS customization in this and other blocks, follow the instruc
 | `orderByButton`                       |
 | `orderByDropdown`                     |
 | `orderByOptionItem`                   |
+| `orderByOptionItem--selected`         |
 | `orderByOptionsContainer`             |
 | `orderByText`                         |
 | `orderBy`                             |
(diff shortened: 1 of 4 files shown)


react/components/SelectionListItem.js
```diff
@@ -19,7 +19,7 @@ const SelectionListItem = ({ option, onItemClick, selected }) => {
   return (
     <button
       className={classNames(
-        handles.orderByOptionItem,
+        applyModifiers(handles.orderByOptionItem, selected && 'selected'),
```
iaronaraujo, line 22: While this does work and is a very clever way of using `applyModifiers`, it wasn't created to be used without passing a valid string or string[] as the second argument:

image

I think it would be better to check `selected` previously so you only use `applyModifiers` if its value is `true`!

BeatrizMiranda: @iaronaraujo take a look if it is better now

iaronaraujo: It is! Thank you!

iaronaraujo:
@BeatrizMiranda approved! Just remember to update the snapshots so the tests can work again!

react/components/SelectionListItem.js
```diff
@@ -21,6 +21,7 @@ const SelectionListItem = ({ option, onItemClick, selected }) => {
       className={classNames(
         handles.orderByOptionItem,
         highlight,
+        selected && handles.orderByOptionSelected,
```
Klynger, line 24: I think the `applyModifiers` is the right way, you just have to pass an empty string if `selected` is `false`

```suggestion
        applyModifiers(handles.orderByOptionItem, selected ? 'selected : ''),
```

docs/README.md
```diff
@@ -392,6 +392,7 @@ In order to apply CSS customization in this and other blocks, follow the instruc
 | `orderByButton`                       |
 | `orderByDropdown`                     |
 | `orderByOptionItem`                   |
+| `orderByOptionSelected`               |
```
Klynger, line 395: If you do the changes remember to change it here too

docs/README.md
```diff
@@ -392,6 +392,7 @@ In order to apply CSS customization in this and other blocks, follow the instruc
 | `orderByButton`                       |
 | `orderByDropdown`                     |
 | `orderByOptionItem`                   |
+| `orderByOptionItem--selected`         |
```
guerreirobeatriz, line 395:
```suggestion
| `orderByOptionItem--{selectedItem}`         |
```

docs/README.md
```diff
@@ -392,6 +392,7 @@ In order to apply CSS customization in this and other blocks, follow the instruc
 | `orderByButton`                       |
 | `orderByDropdown`                     |
 | `orderByOptionItem`                   |
+| `orderByOptionItem--{selectedItem}`         |
```
Klynger, line 395:
```suggestion
| `orderByOptionItem--selected`         |
```

guerreirobeatriz, line 395:
```suggestion
| `orderByOptionItem--selected`         |
```

guerreirobeatriz: So sorry, @BeatrizMiranda. I've made a mistake here, you were right 😄

PR #414, opened 2020-09-03
Feature/filter auto update

hiagolcm:

Add two new props to the filter-navigator

1. `updateOnFilterSelectionOnMobile `

Whether the search will be updated on facet selection (`true`) or not (`false`) when the user is on mobile. The `preventRouteChange` prop must be enabled in order to make this work.

autoupdate

2. `showClearByFilterOnMobile`

Whether a clear button that clear all options in a specific filter should appear beside the filter's name (`true`) or not (`false`)

xJAD9ucu53

#### How to test it?

[Workspace](https://filterautoupdate--storecomponents.myvtex.com/apparel---accessories)


docs/README.md
```diff
@@ -279,14 +279,16 @@ Check out the [**Product Summary documentation**](https://vtex.io/docs/component
...
+| `scrollToTop`                     | `enum`    | Scrolls the page to the top (`auto` or `smooth`) or not (`none`) when selecting a facet.                                                                                                                                                                          | `none`        |
+| `updateOnFilterSelectionOnMobile` | `boolean` | Whether the search will be updated on facet selection (`true`) or not (`false`) when the user is on mobile. The `preventRouteChange` prop must be enabled in order to make this work.                                                                             | `false`       |
+| `showClearByFilterOnMobile`       | `boolean` | Whether a clear button that clear all options in a specific filter should appear beside the filter's name (`true`) or not (`false`).                                                                                                                              | `false`       |
```
guerreirobeatriz, line 291:
```suggestion
| `updateOnFilterSelectionOnMobile` | `boolean` | Whether the search results on mobile should be updated according to filter selection (`true`) or not (`false`). Notice: this prop only works if the `preventRouteChange` prop is declared as `true`.                                                                        | `false`       |
| `showClearByFilterOnMobile`       | `boolean` | Whether a clear button (responsible for erasing all filter options selected by the user) should be displayed alongside the filter name (`true`) or not (`false`).                                                                                                                              | `false`       |
```

guerreirobeatriz (changes requested):

guerreirobeatriz (approved):

react/components/FilterOptionTemplate.js
```diff
@@ -152,7 +169,7 @@ const FilterOptionTemplate = ({
           <ShowMoreFilterButton
             quantity={quantity - FACETS_RENDER_THRESHOLD}
             truncated={truncated}
-            toggleTruncate={() => openTruncated(truncated => !truncated)}
+            toggleTruncate={() => openTruncated((isTruncated) => !isTruncated)}
```
thalytafabrine, line 172: i think this will cause an error 🤔

hiagolcm: well, it should not, I just changed the function parameter name (it was a lint issue. the `no-shadow` one).

thalytafabrine: ahhh sorry, for a moment I thought the function was using the same constant above. we see here a reason to use no-shadow hahah 😆

thalytafabrine (approved):

react/components/FilterOptionTemplate.js
hiagolcm: hahahahah true! Now this rule makes sense

PR #428, opened 2020-09-16
Add two new props to support filtering payment methods

victorhmp:

We recently added new arguments to the `Installments` type in our `search-graphql` schema. These new arguments allow users to filter the returned installment options by payment system names. But currently, we don't expose this feature through `search-result` blocks.

This PR adds two new props:

- `excludedPaymentSystems`
- `includedPaymentSystems`

which are passed as arguments to the `productSearch` query performed by this app.

#### How to test it?

Go to this [Workspace](https://victormiranda--storecomponents.myvtex.com/apparel---accessories/).

#### Related to / Depends on

The arguments cited above come from this PR: https://github.com/vtex-apps/search-graphql/pull/91.

This PR depends on both https://github.com/vtex-apps/store-resources/pull/134 and https://github.com/vtex-apps/store/pull/487.


lbebber (approved):

klzns (approved):

docs/README.md
```diff
@@ -188,7 +188,9 @@ Below you may find all available props to configure your search data (be it by u
...
+| `installmentCriteria`               | `enum`                 | Controls what price should be shown when there are different installments options for it. Possible values are: `MAX_WITHOUT_INTEREST` (displayes the maximum installment option with no interest attached) or `MAX_WITH_INTEREST` (displays the maximum installment option whether it has interest attached or not).                                 | `"MAX_WITHOUT_INTEREST"` |
+| `excludedPaymentSystems`               | `string[]`                 | List of payment systems that should not be taken into account to determine the installments option to be returned. This is only taken into account if `installmentCriteria` is set to `MAX_WITHOUT_INTEREST` or `MAX_WITH_INTEREST`.                                  | `undefined` |
+| `includedPaymentSystems`               | `string[]`                 | List of payment systems that should be taken into account to determine the installments option to be returned. This is only taken into account if `installmentCriteria` is set to `MAX_WITHOUT_INTEREST` or `MAX_WITH_INTEREST`. If this is not set, **all** payment systems are included.                                  | `undefined` |
```
guerreirobeatriz, line 193:
```suggestion
| `installmentCriteria`               | `enum`                 | Controls which price should be shown when there are different installments options for it. Possible values are: `MAX_WITHOUT_INTEREST` (displays the maximum installment option with no interest attached to it) or `MAX_WITH_INTEREST` (displays the maximum installment option whether it has interest attached to it or not).                                 | `"MAX_WITHOUT_INTEREST"` |
| `excludedPaymentSystems`               | `[string]`                 | List of payment systems that should *not* be taken into account when displaying the installment options to users. Caution: this prop configuration only works in scenarios where the `installmentCriteria` prop was also declared. In case it was not, all available payment systems will be displayed regardless.   | `undefined` |
| `includedPaymentSystems`               | `[string]`                 | List of payment systems that should be taken into account when displaying the installment options to users. Caution: this prop configuration only works in scenarios where the `installmentCriteria` prop was also declared. In case it was not, all available payment systems will be displayed regardless.                                  | `undefined` |
```

guerreirobeatriz (changes requested):

guerreirobeatriz (approved):

PR #434, opened 2020-10-09
Fix clear button

thalytafabrine:

there are two problems with the current clear filters button:

- if the initial search includes a filter that is not a category, department or collection, that filter will be removed if you clear the filters.
- if `preventRouteChange` is `false`, the button clears all filters, regardless of type, and consequently leaves the search page.

#### How to test it?

[Before](https://alssports.myvtex.com/clothing/jackets/Mens?map=c,c,specificationFilter_7721)
[After](https://mobileclear--alssports.myvtex.com/clothing/jackets/Mens?map=c,c,specificationFilter_7721&__device=phone)
[After - without `preventRouteChange`](https://mobilewithoutprevent--alssports.myvtex.com/clothing/jackets/Mens?map=c,c,specificationFilter_7721)

- on mobile, select a filter and apply.
- open the filter sidebar, and click the clear button
- only the applied filter must be cleaned, and the search must return to the initial state

Before:
beforeprevent

After:
afterprevent

- initially i tried to save the initial state in a context within the search-result, however this does not work without `preventRouteChange` since the initial state was lost whenever the route changed. for this reason, and because it would be strange to put this search info in a context outside the search, the option that made most sense was to put it in the `window` object. however, if you have an alternative that seems more appropriate, we can change this. :)

### DIFF
--- a/react/SearchResultFlexible.js
+++ b/react/SearchResultFlexible.js
@@ -62,6 +62,12 @@ const SearchResultFlexible = ({
   thresholdForFacetSearch,
   lazyItemsRemaining,
 }) => {
+  const facetsFromPathname = pathname =>
+    pathname
+      .toLowerCase()
+      .split('/')
+      .filter(facet => facet)
+
   useEffect(() => {
     const pathname = window.location.pathname
     const initialSearch = window.initialSearchFacets
@@ -72,8 +78,8 @@ const SearchResultFlexible = ({
     }
 
     if (!preventRouteChange) {
-      const initialFacets = initialSearch.pathname.toLowerCase().split('/')
-      const newFacets = pathname.toLowerCase().split('/')
+      const initialFacets = facetsFromPathname(initialSearch.pathname)
+      const newFacets = facetsFromPathname(pathname)
       const totalEqualFacets = newFacets.filter(facet =>
         initialFacets.includes(facet)
       ).length
--- a/react/SearchResultFlexible.js
+++ b/react/SearchResultFlexible.js
@@ -39,6 +39,52 @@ const useShowContentLoader = (searchQuery, dispatch) => {
   }, [dispatch, isLoading, previousLoading])
 }
 
+const facetsFromPathname = pathname =>
+  pathname
+    .toLowerCase()
+    .split('/')
+    .filter(facet => facet)
+
+const useInitialSearch = (preventRouteChange, facets) => {
+  useEffect(() => {
+    const pathname = window.location.pathname
+    const initialSearch = window.initialSearchFacets
+
+    if (!initialSearch && facets) {
+      window.initialSearchFacets = { ...facets, pathname }
+      return
+    }
+
+    if (!preventRouteChange) {
+      const initialFacets = facetsFromPathname(initialSearch.pathname)
+      const newFacets = facetsFromPathname(pathname)
+      const totalEqualFacets = newFacets.filter(facet =>
+        initialFacets.includes(facet)
+      ).length
+
+      if (
+        (newFacets.length < initialFacets.length &&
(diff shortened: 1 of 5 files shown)


hiagolcm (commented):
clear

On [this url](https://mobilewithoutprevent--alssports.myvtex.com/clothing/?map=c), I can't clear after selecting a brand. Is this expected?

thalytafabrine:
> On [this url](https://mobilewithoutprevent--alssports.myvtex.com/clothing/?map=c), I can't clear after selecting a brand. Is this expected?
> 
> clear

Fixed @hiagolcm 😅

hiagolcm (approved):

thalytafabrine:
@vtex-apps/store-framework-devs hi folks, can you take a look at this PR? will change the behavior of cleaning filters so it is important to have your review 🥺 💙

react/SearchResultFlexible.js
```diff
@@ -61,6 +61,50 @@ const SearchResultFlexible = ({
...
+      window.initialSearchFacets = { ...searchQuery.facets, pathname }
+      return
+    }
+
+    if (
+      searchQuery.facets &&
+      initialSearch.pathname !== pathname &&
+      searchQuery.facets.queryArgs.query !== initialSearch.queryArgs.query
+    ) {
+      window.initialSearchFacets = { ...searchQuery.facets, pathname }
```
klzns, line 104: Can we avoid storing this variable on `window`? We can store mutable variables in refs

klzns (changes requested):

react/utils/initialSearchUtils.js
```diff
@@ -0,0 +1,33 @@
+import { PATH_SEPARATOR, MAP_VALUES_SEP } from '../constants'
+
+export const isInitialFacet = (facet, initialSearch) => {
+  if (!initialSearch) {
+    return false
+  }
+
+  const initialFacets = initialSearch.specificationFilters.flatMap(filter =>
```
Klynger, line 8: Do we have a polyfill to `flatMap` in the store? Do you people know anything about that @thalytafabrine @vtex-apps/store-framework-devs ?

react/SearchResultFlexible.js
thalytafabrine: @klzns I tried to use context but it didn't work without the `preventRouteChange`, because the component is remounted and the context is lost. I think that with refs the behavior would be the same 😕

react/utils/initialSearchUtils.js
lucasecdb: we don't, but we can add it [here](https://github.com/vtex/render-server/blob/master/node/utils/assetLoader.ts#L35) as `Array.prototype.flatMap`

thalytafabrine: @Klynger I opened a PR with this: https://github.com/vtex/render-server/pull/693 😅

Klynger (approved):

react/utils/initialSearchUtils.js
Klynger: Great! LGTM

react/SearchResultFlexible.js
klzns: Can we create a custom hook just to make it clearer of the intent of this hook?

klzns (changes requested):

react/SearchResultFlexible.js
thalytafabrine: done @klzns, you can test [here](https://mobileclear--alssports.myvtex.com/clothing/jackets/Mens?map=c,c,specificationFilter_7721&__device=phone)

klzns (approved):

PR #437, opened 2020-10-16
Feature/gallery layout support

icazevedo:

This PR adds the ability to have multiple layout configurations for search page results. They can vary in relation to the number of items displayed per row and which component is being used to show these products. It removes the product-summary restriction, so you can use two different `product-summary` or something entirely different.

It also provides a component that offers the ability to switch between layouts.

It is backwards compatible, so those who are not using gallery-layout yet should see no difference in their galleries.

#### How to test it?

[This workspace](https://icarosearch--storecomponents.myvtex.com/apparel---accessories/) contains two different layouts to be chosen.

[This workspace](https://icarosearch2--storecomponents.myvtex.com/apparel---accessories/) should assert that it is backwards compatible (no changes were made to store-theme).

ezgif com-gif-maker

#### Depends on

[search-page-context#7](https://github.com/vtex-apps/search-page-context/pull/7)


react/Gallery.tsx
```diff
@@ -0,0 +1,46 @@
+import React from 'react'
+import GalleryLayout, { GalleryLayoutPropsWithSlots } from './GalleryLayout'
+//@ts-ignore
```
victorhmp, line 3: Is this because of `.js` extension in `GalleryLegacy`?
Have you tried adding `allowJs` in the `tsconfig` file?

react/Gallery.tsx
```diff
@@ -0,0 +1,46 @@
...
+import GalleryLegacy from './GalleryLegacy'
+
+export type MobileLayoutMode = 'normal' | 'small' | 'inline'
+
+export interface GalleryLegacyProps {
+  lazyItemsRemaining: number
+  maxItemsPerRow: number
+  minItemWidth: number
+  mobileLayoutMode: MobileLayoutMode
+  products: any[]
```
victorhmp, line 13: Can't we type this based on the GraphQL?

react/GalleryLayout.tsx
```diff
@@ -0,0 +1,133 @@
...
+}
+
+export type Slots = Record<string, ComponentType>
+
+export interface GalleryLayoutProps {
+  layouts: LayoutDescription[]
+  lazyItemsRemaining: number
+  products: any[]
+  showingFacets: boolean
+  summary: any
```
victorhmp, line 31: What is this `summary`?

react/GalleryLayout.tsx
```diff
@@ -0,0 +1,133 @@
...
+  slots: Slots
+}
+
+export type GalleryLayoutPropsWithSlots = Omit<GalleryLayoutProps, 'slots'> &
+  Slots
```
victorhmp, line 36: I would add a comment here, this looks a bit confusing at first sight, 'cause it looks like you're removing `slots` and adding it back again 😂

react/GalleryLayout.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+const LAZY_RENDER_THRESHOLD = 2
+
+const CSS_HANDLES = ['gallery']
```
victorhmp, line 40:
```suggestion
const CSS_HANDLES = ['gallery'] as const
```

react/GalleryLayout.tsx
```diff
@@ -0,0 +1,133 @@
...
+    if (galleryLayout) {
+      layoutDescription = layouts.find(layout => layout.name === galleryLayout)
+    }
+
+    return layoutDescription ?? layouts[0]
+  }, [galleryLayout, layouts])
+
+  const itemsPerRow = useResponsiveValue(currentLayoutDescription.itemsPerRow)
+
+  const rows = useMemo(() => {
```
victorhmp, line 69: ⛏️  [Nitpick] Just to avoid having shadowed variable names

```suggestion
  const galleryRows = useMemo(() => {
```

react/GalleryLayoutOption.tsx
```diff
@@ -0,0 +1,52 @@
...
+interface OptionSlot {
+  isActive?: boolean
+}
+
+interface GalleryLayoutOptionProps {
+  Option: ComponentType<OptionSlot>
+  name: string
+}
+
+const CSS_HANDLES = ['galleryLayoutOption']
```
victorhmp, line 19:
```suggestion
const CSS_HANDLES = ['galleryLayoutOption'] as const
```

react/GalleryLayoutSwitcher.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import classNames from 'classnames'
+import { useCssHandles } from 'vtex.css-handles'
+
+const CSS_HANDLES = ['galleryLayoutSwitcher']
```
victorhmp, line 5:
```suggestion
const CSS_HANDLES = ['galleryLayoutSwitcher'] as const
```

react/SearchResultFlexible.js
```diff
@@ -104,6 +105,7 @@ const SearchResultFlexible = ({
   const [state, dispatch] = useSearchPageStateReducer({
     mobileLayout: mobileLayout.mode1,
     showContentLoader: searchQuery.loading,
+    defaultGalleryLayout: defaultGalleryLayout,
```
victorhmp, line 108:
```suggestion
    defaultGalleryLayout,
```

react/components/GalleryLayoutItem.tsx
```diff
@@ -0,0 +1,48 @@
+import React, { useMemo, useCallback, memo, ComponentType } from 'react'
+
+import ProductSummary from 'vtex.product-summary/ProductSummaryCustom'
+import { usePixel } from 'vtex.pixel-manager/PixelContext'
+import { useSearchPage } from 'vtex.search-page-context/SearchPageContext'
+
+interface GalleryLayoutItemProps {
+  GalleryItemComponent: ComponentType
+  item: any
```
victorhmp, line 9: No way to know the type for this?

react/components/SettingsContext.ts
```diff
@@ -0,0 +1,5 @@
+import { Context, createContext } from 'react'
+
+const SettingsContext: Context<any> = createContext({})
```
victorhmp, line 3: This is a `Record<string, unknown>`, right? Or something like that

react/typings/vtex.css-handles.d.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  export const useCssHandles: <T extends CssHandlesInput>(
+    handles: T,
+    options: CssHandlesOptions = {}
+  ) => CssHandles<T>
+  export const applyModifiers: (
+    handles: string,
+    modifier: string | string[]
+  ) => string
+}
```
victorhmp, line 19: I don't think you need this anymore! The types that are installed by `vtex setup --typings` are correct

react/typings/vtex.responsive-values.d.ts
```diff
@@ -0,0 +1,40 @@
+declare module 'vtex.responsive-values' {
```
victorhmp, line 1: Not sure if you need this. Try running `vtex setup --typings` to see if these types are needed or not

react/typings/vtex.pixel-manager.d.ts
```diff
@@ -0,0 +1,37 @@
+declare module 'vtex.pixel-manager/PixelContext' {
```
victorhmp, line 1: I'm positive you don't need these definitions too

victorhmp (changes requested):

react/Gallery.tsx
icazevedo: Yes, it is. Did not try that, thank you!

react/GalleryLayout.tsx
icazevedo: It is a set of product-summary props. I didn't know how to type it, especially because the legacy gallery had it as type `any` in propTypes.

react/GalleryLayout.tsx
icazevedo: hahahahah agreed!

react/typings/vtex.css-handles.d.ts
icazevedo: I wasn't aware of this command. Thanks!

react/components/SettingsContext.ts
icazevedo: Right! I will change that :)

react/GalleryLayoutOption.tsx
```diff
@@ -0,0 +1,52 @@
...
+      type: SWITCH_GALLERY_LAYOUT_TYPE,
+      args: { galleryLayout: name },
+    })
+  }
+
+  return (
+    <button
+      className={`${handles.galleryLayoutOption} grow dib br-100 pa2 mr2 ml2 bw0 pointer outline-0 bg-transparent`}
+      onKeyDown={() => handleOptionClick()}
+      onClick={() => handleOptionClick()}
```
Klynger, line 42:
```suggestion
      onKeyDown={handleOptionClick}
      onClick={handleOptionClick}
```

react/GalleryLayoutOption.tsx
```diff
@@ -0,0 +1,52 @@
...
+      aria-controls={'layout-option-items'}
+      aria-label={`Option for ${name} layout`}
+      data-testid="galleryLayoutOption"
```
Klynger, line 45: Do you really need this?

react/GalleryLayout.tsx
```diff
@@ -0,0 +1,133 @@
...
+    const rows = []
+
+    let i = 0
+
+    while (i * itemsPerRow < products.length) {
```
Klynger, line 74: I don't remember the last time I saw a `while` in a frontend code, but good tho haha

react/GalleryLayout.tsx
```diff
@@ -0,0 +1,133 @@
...
+      <div className={galleryClasses}>
+        {rows.map((rowProducts, index) => (
+          <GalleryLayoutRow
+            key={index.toString()}
+            products={rowProducts}
+            lazyRender={!!isLazyRenderEnabled && index >= LAZY_RENDER_THRESHOLD}
+            summary={summary}
+            displayMode={'normal'}
+            itemsPerRow={itemsPerRow}
+            GalleryItemComponent={slots[currentLayoutDescription.component]}
```
Klynger, line 112: What would happen if the user doesn't pass this array? Or if it has `null` types?

react/Gallery.tsx
```diff
@@ -0,0 +1,46 @@
...
+  GalleryLegacyProps | GalleryLayoutPropsWithSlots
+> = props => {
+  if ('layouts' in props && props.layouts.length > 0) {
+    const {
+      layouts,
+      lazyItemsRemaining,
+      products,
+      showingFacets,
+      summary,
+      ...slots
```
Klynger, line 28: This seems a bit strange, if the user passes something else that there isn't a slot it would add to this object as a slot. But I don't know if there is some other way to do this

react/GalleryLayout.tsx
```diff
@@ -0,0 +1,133 @@
...
+  return (
+    <ProductListProvider listName={trackingId}>
+      <div className={galleryClasses}>
+        {rows.map((rowProducts, index) => (
+          <GalleryLayoutRow
+            key={index.toString()}
+            products={rowProducts}
+            lazyRender={!!isLazyRenderEnabled && index >= LAZY_RENDER_THRESHOLD}
+            summary={summary}
+            displayMode={'normal'}
```
Klynger, line 110:
```suggestion
            displayMode="normal"
```

react/GalleryLayout.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+  const isLazyRenderEnabled = getSettings('vtex.store')
+    ?.enableSearchRenderingOptimization
+
+  return (
+    <ProductListProvider listName={trackingId}>
+      <div className={galleryClasses}>
+        {rows.map((rowProducts, index) => (
+          <GalleryLayoutRow
+            key={index.toString()}
```
Klynger, line 106: I think you don't need this, but your choice to change
```suggestion
            key={index}
```

react/components/GalleryLayoutItem.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const handleClick = useCallback(() => {
+    push({ event: 'productClick', product, query })
+  }, [product, query, push])
+
+  return (
+    <GalleryItemComponent
+      {...summary}
+      product={product}
+      displayMode={displayMode}
+      actionOnClick={handleClick}
```
Klynger, line 43: ```actionOnClick```

This prop has a really strange name. But I believe this is out of your control, right?

Klynger (changes requested):
This renaming of the file is a breaking change. But I don't know if this really matters cc @klzns @victorhmp @kaisermann

`react/Gallery.js → react/GalleryLegacy.js `

klzns (changes requested):
Still reviewing, submitting now so I can have more time to review

react/GalleryLayout.tsx
klzns: Couldn't we use the name of the gallery-layout here?

react/GalleryLayout.tsx
```diff
@@ -0,0 +1,133 @@
...
+  showingFacets,
+  summary,
+  slots,
+}) => {
+  const { trackingId = 'Search result' } = useContext(SettingsContext) || {}
+  const handles = useCssHandles(CSS_HANDLES)
+  const { getSettings } = useRuntime()
+  const { galleryLayout } = useSearchPageState()
+
+  const currentLayoutDescription = useMemo(() => {
```
klzns, line 57: I find `description` a bit weird, because it's not really a description is more of a `currentLayoutOption`.

What do you think?

react/GalleryLayoutOption.tsx
klzns: Same question as @Klynger 

`data-testid` should be the last resort when testing, can we use the text inside of it to test?

react/components/GalleryLayoutRow.tsx
```diff
@@ -0,0 +1,68 @@
+import React, { ComponentType, memo } from 'react'
+import { useRenderOnView } from '../hooks/useRenderOnView'
+import { useCssHandles, applyModifiers } from 'vtex.css-handles'
+import classNames from 'classnames'
+
+import GalleryItem from './GalleryLayoutItem'
+
+const CSS_HANDLES = ['galleryItem']
```
klzns, line 8:
```suggestion
const CSS_HANDLES = ['galleryItem'] as const
```

klzns (changes requested):

react/GalleryLayout.tsx
icazevedo: I agree, I'll change that :)

react/components/GalleryLayoutItem.tsx
icazevedo: Yup, it is :'(

react/GalleryLayout.tsx
icazevedo: It breaks :/
What's the best way to react to that?

react/GalleryLayoutOption.tsx
icazevedo: I don't think I do, since we don't test it yet. I based myself on a `slider-layout` component, so I just changed the value.

icazevedo:
> This renaming of the file is a breaking change. But I don't know if this really matters cc @klzns @victorhmp @kaisermann
> 
> `react/Gallery.js → react/GalleryLegacy.js `

I created a `Gallery.tsx` file that returns the GalleryLegacy component if no layouts are provided to it (old behavior). I think that assures it will continue to work as before, right?

react/Gallery.tsx
icazevedo: Builder-hub overrides this config 😓. There's another alternative to that?

react/package.json
```diff
@@ -49,7 +51,30 @@
...
+    "vtex.native-types": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.native-types@0.7.1/public/@types/vtex.native-types",
+    "vtex.pixel-manager": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.pixel-manager@1.4.0/public/@types/vtex.pixel-manager",
+    "vtex.product-list-context": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.product-list-context@0.3.0/public/@types/vtex.product-list-context",
+    "vtex.product-summary": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.product-summary@2.62.0-beta.0/public/@types/vtex.product-summary",
+    "vtex.render-runtime": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.render-runtime@8.122.2/public/@types/vtex.render-runtime",
+    "vtex.responsive-values": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.responsive-values@0.3.0/public/_types/react",
+    "vtex.rich-text": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.rich-text@0.11.2/public/@types/vtex.rich-text",
+    "vtex.search-page-context": "https://icarosearch--storecomponents.myvtex.com/_v/private/typings/linked/v1/vtex.search-page-context@0.0.3+build1603226529/public/@types/vtex.search-page-context",
```
icazevedo, line 68: Reminder: update it when `vtex.search-page-context` is published.

react/typings/vtex.responsive-values.d.ts
icazevedo: I ended up needing it. Running the setup command on `vtex.responsive-values` only gave me an empty `node_modules` folder, no typings.

react/Gallery.tsx
icazevedo: I typed it based on the shape (propType) received by the old gallery component

react/components/GalleryLayoutItem.tsx
icazevedo: Typed :) It was of the product type.

react/GalleryLayout.tsx
```diff
@@ -99,17 +77,17 @@ const GalleryLayout: React.FC<GalleryLayoutProps> = ({
...
 
   return (
-    <ProductListProvider listName={trackingId}>
+    <ProductListProvider listName={trackingId as string}>
       <div className={galleryClasses}>
-        {rows.map((rowProducts, index) => (
+        {galleryRows.map((rowProducts, index) => (
           <GalleryLayoutRow
-            key={index.toString()}
+            key={`${currentLayoutOption.name}-${index}`}
```
klzns, line 84:
```suggestion
            key={currentLayoutOption.name}
```

react/typings/globals.d.ts
```diff
@@ -0,0 +1,117 @@
+import { ComponentType } from 'react'
+import { MaybeResponsiveInput } from 'vtex.responsive-values'
+
+declare global {
+  interface GalleryLegacyProps {
```
klzns, line 5: If we import the React components from another app the global types are not imported (like we will miss the feature that we are using in vtex.css-handles and other apps)

So we should not use `globals.d.ts`

klzns (changes requested):

react/typings/globals.d.ts
icazevedo: Got it. Should we set these types in the components they belong to, then? Feels kinda bad to mix regular imports with type imports 😔

klzns: You can use `import type { FooBar } from './module'`

react/GalleryLayout.tsx
icazevedo: This removal makes the layout not to switch. It's not a good thing to set multiple elements with the same key, but if that was the case it should break when viewing any layout, not only when switching.

Also, the key changes when switching between layouts, so it shouldn't be a problem.
Maybe React uses these keys to unmount the elements, and since it finds multiple instances with the same keys, it doesn't know what to do and just leaves them there.

react/GalleryLayoutOption.tsx
```diff
@@ -0,0 +1,49 @@
...
+      args: { galleryLayout: name },
+    })
+  }
+
+  return (
+    <button
+      className={`${handles.galleryLayoutOption} grow dib br-100 pa2 mr2 ml2 bw0 pointer outline-0 bg-transparent`}
+      onKeyDown={handleOptionClick}
+      onClick={handleOptionClick}
+      aria-controls={'layout-option-items'}
```
victorhmp, line 41:
```suggestion
      aria-controls="layout-option-items"
```

react/components/GalleryLayoutItem.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const { push } = usePixel()
+  const { searchQuery } = useSearchPage()
+
+  const product = useMemo(
+    () => ProductSummary.mapCatalogProductToProductSummary(item),
+    [item]
+  )
+
+  const query = useMemo(() => {
+    if (searchQuery && searchQuery.variables) {
```
victorhmp, line 30:
```suggestion
    if (searchQuery?.variables) {
```

victorhmp (approved):

Klynger (changes requested):

react/GalleryLayout.tsx
Klynger: You should have a default item (would be probably the product summary or something), or just check if the component was  passed and if it wasn't, console an error to the user if he is in a dev env

react/GalleryLayout.tsx
klzns: Ops, yeah, totally! I understood this component wrong

I thought it was iterating through galleryLayoutOptions

klzns (approved):

Klynger (approved):
LGTM, but it would be incredibly great if you could make accessibility better in the buttons to change the layout because these buttons seem to work as a toolbar and [toolbars have some specifications about how you should navigate inside of them](https://www.w3.org/TR/wai-aria-practices-1.1/#kbd_general_within). But I think this is up to you and the store framework team to decide the priority of this (and I think this should be done in another PR anyway)

docs/README.md
```diff
@@ -263,24 +264,113 @@ Structure the `search-result-layout` or the `search-result-layout.customQuery`,
...
+
+- **`layouts` object:** 
+
+| Prop name   | Type     | Description                                                             | Default value   |
+| :---------: | :------: | :---------------------------------------------------------------------: | :-------------: |
+| `name`   | `string` | ![https://img.shields.io/badge/-Mandatory-red](https://img.shields.io/badge/-Mandatory-red) Layout name. This value must be unique i.e. not equal to other layout names declared in the `gallery` block. | `undefined` |
+| `component`   | `string` | ![https://img.shields.io/badge/-Mandatory-red](https://img.shields.io/badge/-Mandatory-red) Name of the [Slot](https://vtex.io/docs/concepts/slots) that will declare the block to be rendered in this layout. The same Slot name used as this prop's value must be declared afterwards as a prop (in the `gallery` block) whose value must be the desired block name. Caution: the `component`  prop's value must be equal to the Slot name responsible for declaring the desired block, thus being PascalCased, not to the desired block name itself. Check out the example below in order to understand the underlying logic behind this prop. | `undefined` |
```
guerreirobeatriz, line 294: `thus being PascalCased` -> I didn't understand this part.

guerreirobeatriz: I have a question: how users should know the slot name to be filled in the `component` prop? In the example below, we have `WholeLineSummary` and `ManyByLineSummary`. How should they know about these names? We do not explain this in the Slots documentation.

docs/README.md
```diff
@@ -263,24 +264,113 @@ Structure the `search-result-layout` or the `search-result-layout.customQuery`,
...
+        }
+      ],
+      "WholeLineSummary": "product-summary.shelf#whole",
+      "ManyByLineSummary": "product-summary.shelf#many"
+    }
+  }
+}
+```
+
+#### `gallery-layout-switcher` block
```
guerreirobeatriz, line 333: We should add this block to the app's block list (at the beginning of the doc).

docs/README.md
```diff
@@ -263,24 +264,113 @@ Structure the `search-result-layout` or the `search-result-layout.customQuery`,
...
+
+Allows users to shift between the available `gallery`'s layouts. 
 
-Therefore, don't forget to check out the [Flex Layout documentation](https://vtex.io/docs/apps/layout-blocks/vtex.flex-layout@0.14.0) for more on how to configure your search results page. 
+It receives no props and expects as child the `gallery-layout-option` block described below.
 
-Below you can find the existing props for each of the blocks, in addition to the the rules that govern them. 
+#### `gallery-layout-option` block
```
guerreirobeatriz, line 339: We should add this block to the app's block list (at the beginning of the doc).

guerreirobeatriz: I didn't fully understand this block functionality, therefore I couldn't properly understand as well its props.  What does it do on the UI?

guerreirobeatriz (changes requested):

docs/README.md
icazevedo: `being PascalCased` means that the first letter of each word in a compound word is capitalized. Slots' names must be PascalCased by definition, if not, they won't work.

About the second comment, the Slot name is up to the users to define. They can give it any name they want, but it is required that the `component` prop have the same name as the corresponding Slot that they want to display for that layout.

In the example, we could've used any PascalCased name for the Slot, such as `EntireLineSum`, and then define a layout that receives "EntireLineSum" as the value for the component prop.

docs/README.md
icazevedo: This block is the button used to switch to a given layout. In the description of this PR, the GIF shows me clicking on it to navigate between layouts.

guerreirobeatriz (approved):

PR #439, opened 2020-10-19
Feature/category links

MatheusR42:
This pull request is related to this issue: https://github.com/vtex-apps/store-discussion/issues/371

Today the category links are divs with role="link". It works fine but I think that would be best to use just `<a>` tags as suggested in Mozilla docs: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_link_role and by a SEO team of one of my clients.

With tags we also will improve the experience because the user will can open link in a new tab

#### How to test it?
You can test in this workspace: 

[Workspace](https://mathias1--compracerta.myvtex.com/eletrodomesticos)

There is no visual changes
image

I think that this change may affect some stores because I changed a `<div>` tag to a `<a>` tag. To mantain the compatibility I add the "db no-underline" classes and did not change the css handle. One alternative that I think is to enable this change by prop, but I really do not think that is a prop that makes sense since the  `<a>` is clearly the recommended option.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -282,12 +282,11 @@ Check out the [**Product Summary documentation**](https://vtex.io/docs/component
 
 - **`filter-navigator.v3` block**
 
+
 | Prop name | Type                      | Description                                                                                       | Default value |
 | --------- | ------------------------- | ------------------------------------------------------------------------------------------------- | ------------- |
-
 | `categoryFiltersMode`  | `Enum` | Whether the Category Filter should add href (`href`) or not (`default`). By default the filter uses divs with role="link" you may change to href to improve store Link building. | `default`  |
 | `layout`  | `Enum` | Whether the Filter Navigator layout should be responsive (`responsive`) or not (`desktop`). You may use `desktop` when the Filter Navigator was configured to be displayed in a [drawer](https://vtex.io/docs/components/content-blocks/vtex.store-drawer@0.9.0). | `responsive`  |
-
 | `maxItemsDepartment` | `number`                 | Maximum number of departments to be displayed before the See More button is triggered.          | `8`             |
 | `maxItemsCategory`   | `number`                 | Maximum number of category items to be displayed before the See More button is triggered.     | `8`             |
 | `initiallyCollapsed` | `Boolean` | Makes the search filters start out collapsed (`true`) or open (`false`). | `false` |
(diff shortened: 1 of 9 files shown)


klzns (changes requested):
Thanks for the PR!

>I think that this change may affect some stores because I changed a <div> tag to a <a> tag. To mantain the compatibility I add the "db no-underline" classes and did not change the css handle. One alternative that I think is to enable this change by prop, but I really do not think that is a prop that makes sense since the <a> is clearly the recommended option.

Unfortunately, there's no way we can guarantee that will not break a store, because of that this should be under a prop.

Can make that change and add this new prop to the `docs/README.md`? 

Also it's missing adding to the changelog.

MatheusR42:
@klzns prop added :)

Code to enable this feature
```
"filter-navigator.v3": {
    "props": {
      "categoryFiltersMode": "href"
    }
  }
```

Workspace with feature enabled: https://mathias1--compracerta.myvtex.com/eletrodomesticos

docs/README.md
```diff
@@ -284,6 +284,7 @@ Check out the [**Product Summary documentation**](https://vtex.io/docs/component
 
 | Prop name | Type                      | Description                                                                                       | Default value |
 | --------- | ------------------------- | ------------------------------------------------------------------------------------------------- | ------------- |
+| `categoryFiltersMode`  | `Enum` | Whether the Category Filters should use href (`href`) or not (`default`). By default the filter uses divs with role="link" you may change to href to improve store Link Building. | `default`  |
```
guerreirobeatriz, line 287: I didn't understand the meaning behind _You may change to href to improve store Link Building._ Could you please clarify this prop description for me? :)

despossivel: @MatheusR42 is not around now, but he told me that this can improve the google shop index, I don't know if that is clear to you

MatheusR42: @guerreirobeatriz , Link Building is a technique to improve website SEO. In short when many pages link to a page "X" the page "X" will improve rank score. With the "href" feature enabled we will make every category page to link  to child categories so all categories will improve its page rank.

guerreirobeatriz (approved):

klzns (approved):

thalytafabrine (approved):

PR #441, opened 2020-10-23
Fix search title with percentage

thalytafabrine:

When the search title had a percentage symbol (`%`) the SearchTitle component was not displayed due to an error when trying to run `decodeURI(title)`

[Issue](https://github.com/vtex-apps/store-discussion/issues/442)

#### How to test it?

[Before](https://acctglobal.myvtex.com/141?map=productClusterIds)

[Workspace](https://raissateste--acctglobal.myvtex.com/141?map=productClusterIds)

### DIFF
--- a/react/SearchTitle.js
+++ b/react/SearchTitle.js
@@ -58,14 +58,21 @@ const SearchTitle = props => {
 
   const title =
     index >= 0 ? breadcrumbName(index, breadcrumb) : getLastName(breadcrumb)
+
+  const decodedTitle = useMemo(() => {
+    try {
+      return decodeURI(title)
+    } catch {
+      return title
+    }
+  }, [title])
+
   if (!title) {
     return null
   }
 
   return (
-    <h1 className={classNames(wrapperClass, 't-heading-1')}>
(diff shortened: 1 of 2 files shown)


klzns (approved):

react/SearchTitle.js
```diff
@@ -58,14 +58,21 @@ const SearchTitle = props => {
...
+
+  const decodedTitle = useMemo(() => {
+    try {
+      return decodeURI(title)
+    } catch {
+      return title
+    }
+  }, [title])
+
   if (!title) {
```
icazevedo, line 70: I think this check should be done before the `useMemo` hook. Calling `decodeURI` on `null` or `undefined` returns a string, so this makes this condition unreachable. It also deviates from the previous behavior.

icazevedo (changes requested):

react/SearchTitle.js
icazevedo: Ooops, my bad! I thought you were using decodedTitle 😅 My bad!

icazevedo (approved):

PR #443, opened 2020-10-28
Feature/slider inputs

hiagolcm:

Add text input to the price range component.

#### How to test it?

[Workspace](https://sliderinputs--storecomponents.myvtex.com/apparel---accessories/eyeglasses/)

Desktop:
image

Mobile>
image

#### Related to / Depends on

https://github.com/vtex/styleguide/pull/1343

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -422,7 +422,7 @@ Renders a filter selector for the fetched results.
 | `navigationTypeOnMobile` | `Enum`    | Defines how mobile users should navigate on the filter selector component. Possible values are: `page` (only one list of options can be seen at a time) or `collapsible` (all lists of options can be seen at the same time).                                    | `page`        |
 | `updateOnFilterSelectionOnMobile` | `boolean` | Whether the search results on mobile should be updated according to filter selection (`true`) or not (`false`). Notice: this prop only works if the `preventRouteChange` prop is declared as `true`.                                                                        | `false`       |
 | `showClearByFilter`       | `boolean` | Whether a clear button (responsible for erasing all filter options selected by the user) should be displayed alongside the filter name (`true`) or not (`false`).                                                                                                                              | `false`       |
-| `priceRangeLayout` | `"default"` or `inputAndSlider` | whether a text field should appear to enter the price range (`"inputAndSlider"`), or not (`"default"`). | `"default"` |
+| `priceRangeLayout` | `enum` | Whether a text field to enter the desired price range should be displayed  (`inputAndSlider`) or not (`default`). | `default` |
 
 #### `order-by` block
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -422,6 +422,7 @@ Renders a filter selector for the fetched results.
 | `navigationTypeOnMobile` | `Enum`    | Defines how mobile users should navigate on the filter selector component. Possible values are: `page` (only one list of options can be seen at a time) or `collapsible` (all lists of options can be seen at the same time).                                    | `page`        |
 | `updateOnFilterSelectionOnMobile` | `boolean` | Whether the search results on mobile should be updated according to filter selection (`true`) or not (`false`). Notice: this prop only works if the `preventRouteChange` prop is declared as `true`.                                                                        | `false`       |
 | `showClearByFilter`       | `boolean` | Whether a clear button (responsible for erasing all filter options selected by the user) should be displayed alongside the filter name (`true`) or not (`false`).                                                                                                                              | `false`       |
+| `priceRangeLayout` | `"default"` or `inputAndSlider` | whether a text field should appear to enter the price range (`"inputAndSlider"`), or not (`"default"`). | `"default"` |
 
 #### `order-by` block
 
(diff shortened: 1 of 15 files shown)


docs/README.md
```diff
@@ -292,6 +292,7 @@ Check out the [**Product Summary documentation**](https://vtex.io/docs/component
 | `filtersTitleHtmlTag` | `string` | HTML tag for the filter's title. | `h5` |
 | `scrollToTop` | `enum` | Scrolls the page to the top (`auto` or `smooth`) or not (`none`) when selecting a facet. | `none` |
 | `truncateFilters` | `boolean` | Whether a filter selector with more than 10 filter options should shorten the list and display a `See more` button (`true`) or not (`false`). | `false` |
+| `priceRangeLayout` | `"default"` or `inputAndSlider` | whether a text field should appear to enter the price range (`"inputAndSlider"`), or not (`"default"`). | `"default"` |
```
guerreirobeatriz, line 295:
```suggestion
| `priceRangeLayout` | `enum` | Whether a text field to enter the desired price range should be displayed  (`inputAndSlider`) or not (`default`). | `default` |
```

guerreirobeatriz: Wouldn't it be better if this prop was boolean? `true` for enabling the text field and `false` for not displaying it.

guerreirobeatriz (changes requested):

docs/README.md
hiagolcm: We are facing problems related to boolean props. It turns out they are not future-proof 😞 .
For example, if someone creates another priceRangeLayout in the future, it will require a new prop that would conflict with this one.

guerreirobeatriz (approved):

react/FilterNavigator.js
```diff
@@ -81,6 +81,7 @@ const FilterNavigator = ({
   totalProductsOnMobile = 'hide',
   fullWidthOnMobile = false,
   navigationTypeOnMobile = 'page',
+  priceRangeLayout = 'default',
```
icazevedo, line 84:
```suggestion
  priceRangeLayout = 'slider',
```

I think it would be more clear if the default option had a name representing its actual behavior

docs/README.md
```diff
@@ -297,6 +297,7 @@ Check out the [**Product Summary documentation**](https://vtex.io/docs/component
 | `totalProductsOnMobile` | `enum` | Whether the Filter Navigator should display the total number of products on mobile devices  (`show`) or not (`hide`). | `hide` |
 | `fullWidthOnMobile`      | `boolean` | Whether the `filter-navigator.v3` will be rendered on mobile using the screen full width (`true`) or not (`false`).                                                                                                                                               | `false`       |
 | `navigationTypeOnMobile` | `Enum`    | Defines how mobile users should navigate on the filter selector component. Possible values are: `page` (only one list of options can be seen at a time) or `collapsible` (all lists of options can be seen at the same time).                                    | `page`        |
+| `priceRangeLayout` | `enum` | Whether a text field to enter the desired price range should be displayed  (`inputAndSlider`) or not (`default`). | `default` |
```
icazevedo, line 300:
```suggestion
| `priceRangeLayout` | `enum` | Whether a text field to enter the desired price range should be displayed  (`inputAndSlider`) or not (`slider`). | `slider` |
```

react/components/PriceRange/PriceRangeInput.js
```diff
@@ -0,0 +1,109 @@
...
+      }))
+    } else {
+      setValues(currentValues => ({
+        ...currentValues,
+        right: parseFloat(value),
+      }))
+    }
+  }
+
+  const handleSubmit = () => {
```
icazevedo, line 46: I'm being able to submit left values bigger than the maximum value and right values smaller than the minimum value. It needs to check those cases as well.

icazevedo (changes requested):

react/FilterNavigator.js
hiagolcm: Agree! It is clearer now.

react/components/PriceRange/PriceRangeInput.js
hiagolcm: I didn't notice that! Now it is fixed!

react/components/PriceRange/PriceRangeInput.js
```diff
@@ -0,0 +1,101 @@
...
+    const { left: leftValue, right: rightValue } = values
+    const left = Math.max(Math.min(leftValue, rightValue), min)
+    const right = Math.min(Math.max(leftValue, rightValue), max)
```
icazevedo, line 48: 👌

icazevedo (approved):
The code LGTM, but danger is signaling some lint errors on the  `react/FilterNavigatorFlexible.js` file. I can't tell if it's a false positive or not, but it's better to check that before merging :)

Also, I couldn't find the slider/slider input on the workspace you provided

camilavcoutinho (approved):

hiagolcm:
> Also, I couldn't find the slider/slider input on the workspace you provided

Sorry it was my fault. I updated the URL 😉

icazevedo:
> > Also, I couldn't find the slider/slider input on the workspace you provided
> 
> @icazevedo Sorry it was my fault. I updated the URL 😉

Awesome! On mobile, I've noticed that when I click "clear" it doesn't restore the default values on the input/slider. Is this intentional?

hiagolcm:
> Awesome! On mobile, I've noticed that when I click "clear" it doesn't restore the default values on the input/slider. Is this intentional?

hiagolcm:
> > > Also, I couldn't find the slider/slider input on the workspace you provided
> > 
> > 
> > @icazevedo Sorry it was my fault. I updated the URL 😉
> 
> Awesome! On mobile, I've noticed that when I click "clear" it doesn't restore the default values on the input/slider. Is this intentional?

This is a known issue 😢 . We have a card to deal with this problem. It was already happening before this PR 😢 

After [this PR](https://github.com/vtex/styleguide/pull/1365) I think it will be easy to fix it. I'll check with my team if we can prioritize this bug since it is so easy to fix

thalytafabrine (approved):

PR #410, opened 2020-08-26
Feature/filter navigator improvements

hiagolcm:
#### What problem is this solving?

This PR adds two new props to the filter-navigator.

1. `fullWidthOnMobile`. Now the `Sidebar` can occupy the whole window on Mobile.

fullwidth

2. `navigationTypeOnMobile`. Now a new navigation type is available on mobile where the filters are collapsible.

navigationType

It also fixes a bug where the scroll was not working on the filter selection. 

Here is how the bug looks like:
bug

and the fix:
bugfix

#### How to test it?

[Workspace](https://filterimprovements--storecomponents.myvtex.com/apparel---accessories/)

PR #448, opened 2020-11-04
Add fetch optimization for collapsible filters on mobile

thalytafabrine:

Make `enableFiltersFetchOptimization` work on mobile.

#### How to test it?

[Collapsible](https://thalyta--storecomponents.myvtex.com/apparel---accessories/clothing)

[Page](https://thalyta1--storecomponents.myvtex.com/apparel---accessories/clothing)

there are two different behaviors:
1 - when using filters in `collapsible` mode, filters must display a maximum of 10 items and fetch by clicking on `see more`.
2 - when using filters in `page` mode, fetching must be done when clicking on the filter, showing all facets when opening.

Collapsible:

collapsiblefetch

Page:

pagefetch

#### Related to / Depends on

Depends on: https://github.com/vtex-apps/search-result/pull/410

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Changed
+- `enableFiltersFetchOptimization` now working on mobile.
+
 ## [3.85.4] - 2020-11-13
 
 ### Fixed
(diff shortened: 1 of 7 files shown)


icazevedo (approved):
LGTM. I've seen that `enableFiltersFetchOptimization` makes the filter count inaccurate because it counts only the selected filters among the ones that have been fetched already. Since this behavior is present on the desktop also (and is more severe there), I don't think it has anything to do with this issue, but I think we should look into it.

hiagolcm (approved):

guerreirobeatriz (approved):

PR #462, opened 2020-11-30
Bugfix/redirects to home when all facets are unselected

hiagolcm:

https://github.com/vtex-apps/store-discussion/issues/489

The user is redirected to the home when all facets are unselected. To prevent this behavior, I disabled the unselect option when only one facet is selected.

Also, I changed the strategy to decide which search query is the initial one. Instead of using the `window.initialFacets`, I'm using query strings. I made this change for two reasons:
1. `window.initialFacets` failed to predict the initial search query in some edge cases. For example, The selected facet is `A,` and the initial search is also `A`. If you select `B` and then unselect `A`, the initial search will now be B, but it should be A
2. Since `window.initialFacets` is not a state or prop, we cannot track its changes using the React life cycle. It may become a problem in the future.

#### How to test it?

[Workspace](https://hiago--storecomponents.myvtex.com/)

image


klzns (approved):
Great solution!

thalytafabrine (approved):

icazevedo (approved):
Looks good. I've noted that I can still disable the department option on the mobile filters, although it doesn't redirect me to the home page when I apply them. I think this behavior should be addressed in another PR.

PR #468, opened 2020-12-04
Gallery layouts fix and accessibility features

icazevedo:

This PR adds the ability to have multiple layout configurations for search page results. They can vary in relation to the number of items displayed per row and which component is being used to show these products. It removes the product-summary restriction, so you can use two different `product-summary` or something entirely different.

It also provides a component that offers the ability to switch between layouts, now with accessibility built-in. I've used [this guide](https://www.w3.org/TR/wai-aria-practices-1.1/examples/toolbar/toolbar.html) as reference for the accessibility features.

It is backwards compatible, so those who are not using gallery-layout yet should see no difference in their galleries.

#### How to test it?

[This workspace](https://icarosearch--storecomponents.myvtex.com/apparel---accessories/) contains two different layouts to be chosen.

[This workspace](https://icarosearch2--storecomponents.myvtex.com/apparel---accessories/) should assert that it is backwards compatible (no changes were made to store-theme).

[This workspace](https://icarosearch--spaziolusso.myvtex.com/gioielli/anelli) this workspace can be used to test on the store that broke the last time gallery layout was merged.

ezgif com-gif-maker (1)

#### Depends on / Related to

**Depends on** [search-page-context#8](https://github.com/vtex-apps/search-page-context/pull/8)

Related to the previous PR [search-result#437](https://github.com/vtex-apps/search-result/pull/437)


react/package.json
```diff
@@ -29,27 +31,52 @@
...
+    "vtex.native-types": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.native-types@0.7.4/public/@types/vtex.native-types",
+    "vtex.pixel-manager": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.pixel-manager@1.4.0/public/@types/vtex.pixel-manager",
+    "vtex.product-list-context": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.product-list-context@0.3.0/public/@types/vtex.product-list-context",
+    "vtex.product-summary": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.product-summary@2.65.0/public/@types/vtex.product-summary",
+    "vtex.render-runtime": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.render-runtime@8.126.4-beta.1/public/@types/vtex.render-runtime",
+    "vtex.responsive-values": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.responsive-values@0.3.0/public/_types/react",
+    "vtex.rich-text": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.rich-text@0.11.2/public/@types/vtex.rich-text",
+    "vtex.search-page-context": "https://icarosearch--storecomponents.myvtex.com/_v/private/typings/linked/v1/vtex.search-page-context@0.1.0+build1607092293/public/@types/vtex.search-page-context",
```
icazevedo, line 70: Update dependency when [search-page-context#8](https://github.com/vtex-apps/search-page-context/pull/8) is merged

react/GalleryLayoutSwitcher.tsx
```diff
@@ -0,0 +1,92 @@
...
+  )
+
+  const navigateWithKeys = useCallback(
+    (e: React.KeyboardEvent<HTMLDivElement>) => {
+      if (
+        galleryLayouts &&
+        (e.key === 'ArrowLeft' ||
+          e.key === 'ArrowUp' ||
+          e.key === 'ArrowRight' ||
+          e.key === 'ArrowDown')
```
klzns, line 31: Maybe an early return to avoid a big if

Something like

```js
if (!tanana) {
 return
}
// the code that is inside if here
```

react/GalleryLayoutSwitcher.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+        if (e.key === 'ArrowLeft') {
+          newFocusedLayoutIndex = Math.max(0, currentFocusedIndex - 1)
+        } else if (e.key === 'ArrowRight') {
+          newFocusedLayoutIndex = Math.min(
+            currentFocusedIndex + 1,
+            galleryLayouts.length - 1
+          )
+        } else if (e.key === 'ArrowUp') {
+          newFocusedLayoutIndex =
```
klzns, line 54: What is the difference using arrow up and down from left and right?

klzns (changes requested):

react/GalleryLayoutSwitcher.tsx
icazevedo: Left and right keys do nothing when they reach their corners. Up and down loops through the items when that happens.

klzns (approved):

react/package.json
```diff
@@ -60,23 +60,23 @@
...
-    "vtex.native-types": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.native-types@0.7.1/public/@types/vtex.native-types",
+    "vtex.native-types": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.native-types@0.7.4/public/@types/vtex.native-types",
     "vtex.pixel-manager": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.pixel-manager@1.4.0/public/@types/vtex.pixel-manager",
     "vtex.product-list-context": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.product-list-context@0.3.0/public/@types/vtex.product-list-context",
-    "vtex.product-summary": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.product-summary@2.63.0/public/@types/vtex.product-summary",
-    "vtex.render-runtime": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.render-runtime@8.124.3/public/@types/vtex.render-runtime",
+    "vtex.product-summary": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.product-summary@2.65.0/public/@types/vtex.product-summary",
+    "vtex.render-runtime": "http://vtex.vtexassets.com/_v/public/typings/v1/vtex.render-runtime@8.126.4-beta.1/public/@types/vtex.render-runtime",
```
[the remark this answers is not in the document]
klzns: Maybe @salesfelipe did that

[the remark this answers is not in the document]
salesfelipe: I didn't :harold-pain:

react/GalleryLayout.tsx
```diff
@@ -0,0 +1,167 @@
...
+  const currentLayoutOption = useMemo(() => {
+    let layoutOption
+
+    if (selectedGalleryLayout) {
+      layoutOption = layouts.find(
+        (layout) => layout.name === selectedGalleryLayout
+      )
+    } else {
+      console.error(
+        'No default gallery layout defined. Set it in search-result-layout by using the defaultGalleryLayout prop.'
```
vitorflg, line 75: Do we not set a default layout automatically?

vitorflg: ahahah ok, I got it rs

vitorflg (approved):

guerreirobeatriz (approved):

react/Gallery.tsx
```diff
@@ -0,0 +1,107 @@
+import React from 'react'
```
victorhmp, line 1: Careful with the renaming of `Gallery.js` to `GalleryLegacy.js`. People could be using this component through:
`import { Gallery } from 'vtex.search-result'`
In that case, they would now be importing **this** code, and not the legacy one

victorhmp: From this code, it looks fine to me, and I see that you're taking that into account! Pointing this out just to keep in mind :)
I recommend linking this branch in `carrefour` and `motorola`, since they're the most likely to be customizing something with this component.

store/interfaces.json
```diff
@@ -24,9 +24,17 @@
     "allowed": ["shop-review-summary", "sidebar-close-button"]
   },
   "gallery": {
-    "required": ["product-summary"],
+    "allowed": ["product-summary"],
```
victorhmp, line 27: This change _could_ cause issues in some stores. I remember a Slack thread about this a few days or weeks back... is this safe now?

victorhmp (approved):

store/interfaces.json
icazevedo: We discussed it in the last groom. We found out we can define a default block implementation on the `blocks.json` file. That's what I did, so now if there's no explicit gallery definition, we can put a default `product-summary` block there and it will not break :)

victorhmp: Awesome!

react/package.json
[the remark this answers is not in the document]
icazevedo: I've fixed it by setting up the types again. Very strange, I still don't know why that happened.

PR #475, opened 2021-01-07
Pass the map prop to productClick event

vitoria:
I need to know if the `productClick` event was triggered after a full-text search or just by applying filters.

#### How to test it?

Go to the search result page and click on a product, the `productClick` event body should contain the `map` property.

[Workspace](https://waza--storecomponents.myvtex.com/)

Another alternative is to pass different fields for the search terms and for filters, but it could be a breaking change. 😟 

#### Related to / Depends on
New metrics for navigation events in Search and Personalization dashboards.

### DIFF
--- a/react/components/GalleryItem.js
+++ b/react/components/GalleryItem.js
@@ -26,7 +26,9 @@ const GalleryItem = ({ item, displayMode, summary }) => {
     }
   }, [searchQuery])
 
-  const map = useMemo(() => searchQuery?.variables?.map, [searchQuery?.variables?.map])
+  const map = useMemo(() => searchQuery?.variables?.map, [
+    searchQuery?.variables?.map,
+  ])
 
   const handleClick = useCallback(() => {
     push({ event: 'productClick', product, query, map })
--- a/react/components/GalleryItem.js
+++ b/react/components/GalleryItem.js
@@ -21,13 +21,19 @@ const GalleryItem = ({ item, displayMode, summary }) => {
   )
 
   const query = useMemo(() => {
-    if (searchQuery && searchQuery.variables) {
+    if (searchQuery?.variables) {
       return searchQuery.variables.query
     }
   }, [searchQuery])
 
+  const map = useMemo(() => {
+    if (searchQuery?.variables) {
+      return searchQuery.variables.map
+    }
+  }, [searchQuery])
+
   const handleClick = useCallback(() => {
-    push({ event: 'productClick', product, query })
+    push({ event: 'productClick', product, query, map })
   }, [product, query, push])
 
   return (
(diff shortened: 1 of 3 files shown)


react/components/GalleryLayoutItem.tsx
```diff
@@ -32,8 +32,14 @@ const GalleryLayoutItem: React.FC<GalleryLayoutItemProps> = ({
     }
   }, [searchQuery])
 
+  const map = useMemo(() => {
+    if (searchQuery?.variables) {
+      return searchQuery.variables.map
+    }
+  }, [searchQuery])
```
klzns, line 39: I think it's better to use the variable directly so the memo runs fewer times 
```suggestion
  const map = useMemo(
    () => searchQuery?.variables?.map,
    [searchQuery?.variables?.map]
  )
```

react/components/GalleryItem.js
```diff
@@ -21,13 +21,19 @@ const GalleryItem = ({ item, displayMode, summary }) => {
...
+    if (searchQuery?.variables) {
       return searchQuery.variables.query
     }
   }, [searchQuery])
 
+  const map = useMemo(() => {
+    if (searchQuery?.variables) {
+      return searchQuery.variables.map
+    }
+  }, [searchQuery])
```
klzns, line 33: same thing

klzns (changes requested):

react/components/GalleryLayoutItem.tsx
vitoria: Nice, thank you! 😊

react/components/GalleryItem.js
```diff
@@ -21,13 +21,15 @@ const GalleryItem = ({ item, displayMode, summary }) => {
   )
 
   const query = useMemo(() => {
-    if (searchQuery && searchQuery.variables) {
+    if (searchQuery?.variables) {
       return searchQuery.variables.query
     }
   }, [searchQuery])
 
+  const map = useMemo(() => searchQuery?.variables.map, [searchQuery])
```
thalytafabrine, line 29:
```suggestion
  const map = useMemo(() => searchQuery?.variables?.map, [searchQuery?.variables?.map])
```

react/components/GalleryLayoutItem.tsx
```diff
@@ -32,8 +32,10 @@ const GalleryLayoutItem: React.FC<GalleryLayoutItemProps> = ({
     }
   }, [searchQuery])
 
+  const map = useMemo(() => searchQuery?.variables.map, [searchQuery])
```
thalytafabrine, line 35: same here

thalytafabrine (approved):

klzns (approved):

PR #490, opened 2021-02-11
Fix decoded search title

thalytafabrine:

the `decodeURI` function does not work for cases where the encoded character is a slash:
image

#### How to test it?

[Workspace](https://encodeurl--biscoind.myvtex.com/3-48m%2fs?_q=3-48m/s&map=ft)

Before:
image

After:
image

### DIFF
--- a/react/SearchTitle.js
+++ b/react/SearchTitle.js
@@ -1,5 +1,6 @@
 import classNames from 'classnames'
 import React, { useContext, useMemo } from 'react'
+// eslint-disable-next-line no-restricted-imports
 import {
   compose,
   equals,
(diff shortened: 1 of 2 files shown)


react/SearchTitle.js
```diff
@@ -1,5 +1,6 @@
 import classNames from 'classnames'
 import React, { useContext, useMemo } from 'react'
+// eslint-disable-next-line no-restricted-imports
```
icazevedo, line 3: Is this still needed?

icazevedo (approved):
LGTM. I spent some minutes trying to find where did `decodeURIComponent ` come from 😂

react/SearchTitle.js
thalytafabrine: I think so, since the eslint throws an error due to the use of `ramda` 😕

hiagolcm (approved):

PR #492, opened 2021-02-23
Adjust OrderBy when OrderByScoreDESC is selected

italo-aurelio:
I've noticed when I deliberated choose option OrderByScoreDESC there's no effect on store search, then I added this option again and kept the default option as relevance when no option is provided.

When I select orderByField with OrderByScoreDESC nothing happens.

```json
"context": {
      "orderByField": "OrderByScoreDESC",
      "hideUnavailableItems": true,
      "maxItemsPerPage": 24
}
```

image

#### How to test it?

Add option OrderByScoreDESC into context props and check text inside store search result

image

[Workspace](https://task234--tackoftheday.myvtex.com/rider)

Search JSON example:

```json
"context": {
        "orderByField": "OrderByScoreDESC",
        "hideUnavailableItems": true,
        "maxItemsPerPage": 24
      }
```

Before:

image

After:

image

Maybe we can keep how it is by know, but needs to update on docs to developers know how to use or select items by `OrderByScoreDESC` and remove `OrderByScoreDESC ` from possible values.

How we can avoid duplicated options "Relevance" and allow developer to use by default or by selecting `OrderByScoreDESC `?

### DIFF
--- a/react/OrderBy.js
+++ b/react/OrderBy.js
@@ -10,6 +10,10 @@ export const SORT_OPTIONS = [
     label: 'store/ordenation.relevance',
   },
   {
+    value: 'OrderByScoreDESC',
+    label: 'store/ordenation.relevance',
+  },
+  {
     value: 'OrderByTopSaleDESC',
     label: 'store/ordenation.sales',
   },


react/OrderBy.js
```diff
@@ -9,6 +9,10 @@ export const SORT_OPTIONS = [
     value: '',
```
icazevedo, line 9: Maybe a better solution would be to set this value as `OrderByScoreDESC`. Duplicating it is not an option. @victorhmp can you give some input on this?

icazevedo (changes requested):
Thanks for the PR! In fact, that's not the right behavior. The solution isn't adequate either, since it leaves two identical options. Let's wait for @victorhmp's input :)

italo-aurelio:
> Thanks for the PR! In fact, that's not the right behavior. The solution isn't adequate either, since it leaves two identical options. Let's wait for @victorhmp's input :)

Exact, I've pointed this in "Describe alternatives you've considered, if any.". Maybe we can just change on documentation this info, but it stills affecting stores that already use this configuration.

I'm open to better ideas or patterns to follow.

Thanks.

react/OrderBy.js
victorhmp: Sure!
I think the best solution would simply be to not change anything here in `SORT_OPTIONS` and add a "translation" step in the component itself. Something along the lines of

```ts
if (receivedOrderBy === 'OrderByScoreDES') {
  orderBy = ''
}
```

The thing is, this option of `OrderByScoreDESC` used to be our default. But this is also how the Catalog API returns products, except when showing products inside a collection since collections can have custom product orderings. Basically, the ordering people see in search result pages was not affect at all, unless they're seeing a collection's page, but in that case, it changed for the better.

We need to keep this `value: ''` so that we send an empty string as the value for the `orderBy` query variable. This way the API just returns products ordered according to its default, exactly as we want.

icazevedo: Awesome, thanks! @italo-batista, can you do that?

italo-aurelio: Of course, I'll send a new PR with this.

PR #526, opened 2021-07-09
is set so that the search-not-found-layout block is displayed when a …

jairandresdiazp:
…search does not return results or when a category or department path does not return results

is set so that the search-not-found-layout block is displayed when a search does not return results or when a category or department path does not return results

#### How to test it?

https://jomarr--aquamaxxpe.myvtex.com/hola%20coom%20estas?_q=hola%20coom%20estas&map=ft

https://jomarr--aquamaxxpe.myvtex.com/Parrillas/Empotrables

image
image

#### How does this PR make you feel? [:link:](http://giphy.com/)

This helps us to better customize the basic search message since the default block will always be shown in the type /department /category path

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1,7 +1,7 @@
 {
   "name": "search-result",
   "vendor": "vtex",
-  "version": "3.104.0",
+  "version": "3.104.1",
   "title": "VTEX Search Result",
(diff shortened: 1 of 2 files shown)


jairandresdiazp:
falto actulizar el manifest

PR #531, opened 2021-07-19
change from productId to cacheId

hiagolcm:

The intelligent search has a feature called `split by sku`. This is used to show multiple skus on the search-result. Like this

image

Those five products are actually the same product, but different SKUs

2nb427w5d5

Since they are the same product, they have the same `productId`. For this reason, we can't use the `productId` as the component key anymore

#### How to test it?

[Workspace](https://hiago--paylessus.myvtex.com/)

### DIFF
--- a/react/components/GalleryLayoutRow.tsx
+++ b/react/components/GalleryLayoutRow.tsx
@@ -52,10 +52,9 @@ const GalleryLayoutRow: React.FC<GalleryLayoutRowProps> = ({
     <>
       {products.map((product, index) => {
         const absoluteProductIndex = rowIndex * itemsPerRow + index + 1
-
         return (
           <div
-            key={product.productId}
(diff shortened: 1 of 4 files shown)


react/components/GalleryLayoutRow.tsx
```diff
@@ -52,10 +52,9 @@ const GalleryLayoutRow: React.FC<GalleryLayoutRowProps> = ({
     <>
       {products.map((product, index) => {
         const absoluteProductIndex = rowIndex * itemsPerRow + index + 1
-
         return (
           <div
-            key={product.productId}
+            key={product.cacheId}
```
thalytafabrine, line 57: could you test it in a store with the search-resolver@0.x? I think it won't cause any problems but just to make sure 😅

thalytafabrine (approved):

PR #532, opened 2021-07-27
fix: urls on fetch more/fetch previous buttons now includes the map query string

icazevedo:

This PR fixes URLs that were not including the `map` query string when referencing next or previous pages. This is a problem specially on collection pages, since these pages need the `map` property to work. Before, crawlers would be redirected to a 404 page.

#### How to test it?

Look at the Fetch More/Fetch Previous buttons present on the page. This is not the exact same scenario but reflects that the `map` is now being included on the URL when needed. If you remove the `map` from the URL, it doesn't work.

[Workspace](https://icarocnc--storecomponents.myvtex.com/apparel---accessories/home---decor?map=category-1%2Ccategory-1)

Screen Shot 2021-07-27 at 12 24 16

#### Related to / Depends on

Related to https://github.com/vtex-apps/store/pull/529

### DIFF
--- a/react/components/loaders/FetchMoreButton.js
+++ b/react/components/loaders/FetchMoreButton.js
@@ -2,6 +2,7 @@ import React, { Fragment, useState, useEffect } from 'react'
 import { Button } from 'vtex.styleguide'
 import { useCssHandles } from 'vtex.css-handles'
 import { FormattedMessage } from 'react-intl'
+import { useSearchPage } from 'vtex.search-page-context/SearchPageContext'
 
 const CSS_HANDLES = [
   'buttonShowMore',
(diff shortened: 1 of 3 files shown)


react/components/loaders/FetchMoreButton.js
```diff
@@ -23,6 +24,28 @@ const useShowButton = (to, products, loading, recordsFiltered) => {
   return showButton
 }
 
+function shouldNotIncludeMap(map) {
```
igorbrasileiro, line 27: What do you think about reuse this function from the vtex.store?

icazevedo: I considered it, but I thought it was too much hustle since `vtex.store` already depends on `vtex.search-result`. I'm afraid to deal with complicated problems that may arise from such a simple requirement.

victorhmp (approved):

thalytafabrine (approved):

PR #554, opened 2021-11-18
Edu 5216 simulation behavior como default

karenkrieger:

Replaced `skip` values for `default` on the documentation

#### How to test it?

[Workspace](Link goes here!)


docs/README.md
```diff
@@ -143,14 +143,14 @@ Check all props to configure your search data int the table below:
...
 | `skusFilter`           | `enum` | Refines the SKUs returned for each product in the query. The fewer returned SKUs, the more performant your shelf query will be. Available value options: `FIRST_AVAILABLE` (returns only the first available SKU), `ALL_AVAILABLE` (returns all available SKUs), and `ALL` (returns all product's SKUs).                                     | `ALL_AVAILABLE` |
-| `simulationBehavior`     | `enum` | Defines whether the search data will be up-to-date (`default`) or fetched using the Cache (`skip`). You should only use the last option if you prefer faster queries than the most up-to-date prices or inventory.                                                               | `default` |
+| `simulationBehavior`     | `enum` | Defines whether the search data will be up-to-date (`default`) or fetched using the Cache (`default`). You should only use the last option if you prefer faster queries than the most up-to-date prices or inventory.                                                               | `default` |
```
thalytafabrine, line 146:
```suggestion
| `simulationBehavior`     | `enum` | Defines whether the search data will be up-to-date (`default`) or fetched using the Cache (`skip`). You should only use the last option if you prefer faster queries than the most up-to-date prices or inventory.                                                               | `default` |
```

docs/README.md
```diff
@@ -143,14 +143,14 @@ Check all props to configure your search data int the table below:
...
 | `includedPaymentSystems`               | `string`                 | List of payment systems that should be considered when displaying the installment options to users. This prop configuration only works if the `installmentCriteria` prop was also declared. In case it was not, all available payment systems will be displayed regardless.                                  | `undefined` |
 
 
 
-> ⚠️ When the `simulationBehavior` prop is set as `skip`, it defines that the search data should only be fetched using the store's cache. This may impact the content displayed on store pages since the cache storage changes according to user interaction in each page.
+> ⚠️ When the `simulationBehavior` prop is set as `default`, it defines that the search data should only be fetched using the store's cache. This may impact the content displayed on store pages since the cache storage changes according to user interaction in each page.
```
thalytafabrine, line 153:
```suggestion
> ⚠️ When the `simulationBehavior` prop is set as `skip`, it defines that the search data should only be fetched using the store's cache. This may impact the content displayed on store pages since the cache storage changes according to user interaction in each page.
```
skip here is right, we just needed to change the examples

thalytafabrine (changes requested):

PedroAntunesCosta (approved):
LGTM

thalytafabrine (approved):

PR #562, opened 2021-12-14
Fix/orderby behaviour on custom query

gabrielHosino:

When we use the SearchResultLayoutCustomQuery block, changing the orderby select does not change the order of the products shown. This PR seeks to correct that, by adding a check if the order key exists in the querystring of the URL. 

#### How to test it?

You can use this [Workspace](https://affiliates--gabrielhosinoio.myvtex.com/affiliates/lojasa) with the new solution working. That link will take you to a new custom page the FPA BR team is working on. There, we use the SearchResultLayoutCustomQuery 

https://user-images.githubusercontent.com/8519076/145996367-b97260c3-95ba-4f92-8a4d-daf8b425efe9.mp4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fix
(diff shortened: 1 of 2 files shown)


hiagolcm (approved):
Simple and clean. thanks for the contribution

PR #589, opened 2022-06-09
Types notation and object clone fixes

ganobrega:
- Changed the name of interface `Filters` to `Filter`
- Changed the name of interface `Facets` to `Facet`
- Changed the object clone method

1. Typescript code improvements
2. Object Clone improvements
   In this example, we can see what happens when using a clone of object with spread operator: 
Captura de Tela 2022-06-09 às 14 55 48

When using the spread operator, the variable `filterCopy` points to the same object `filter` in memory. Not making sense to use `filterCopy`.

In this example, we can see what happens when using a clone of object with `JSON.parse` and `JSON.stringify`:
Captura de Tela 2022-06-09 às 14 55 59

The object in `filterCopy` is really a copy of `filter`, and we can change securely.

#### How to test it?
- `vtex test`

### DIFF
--- a/react/utils/sortFilterValues.ts
+++ b/react/utils/sortFilterValues.ts
@@ -1,6 +1,6 @@
 import { SPEC_FILTERS, ASC, DESC } from '../constants'
 
-interface Filters {
+interface Filter {
   key: string
   quantity: number
   title: string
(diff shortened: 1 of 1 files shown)


thalytafabrine (commented):
can you update the changelog?

PR #591, opened 2022-06-22
replace react-outside-click-handler

hiagolcm:

[`react-outside-click-handler` is a large lib and the maintainers don't have plans to decrease it](https://github.com/airbnb/react-outside-click-handler/issues/24)

This PR replaces the react-outside-click-handler with a 1st party component

#### How to test it?

[Workspace](https://hiago--storecomponents.myvtex.com/)

Open the filter and close it by clicking outside of it


thalytafabrine (commented):
I think your workspace is outdated, but the code looks good to me 🙃 

image

PR #592, opened 2022-07-08
Add structured data

polishq:

In the past, IO search pages imported the `ProductList` component from `vtex.structured-data` so that schema.org LD+JSON data would be rendered to the page. Example [here](https://github.com/vtex-apps/store/blob/431dcd37736f670f620bbd85af11db262d22c8a2/react/SearchWrapper.tsx#L440). This is very helpful for SEO. 

This PR attempts to do the same, importing the `ProductList` component and rendering it as part of this app's `Gallery` component. 

#### How to test it?

Linked here: https://arthur--budgetgolf.myvtex.com/category/mens-golf-vests

and you can see the results in Google's Rich Results test: https://search.google.com/test/rich-results/result/r%2Fproduct?id=CGCKmWDfO399fO2z6LV-Qg


.prettierrc
```diff
@@ -1 +1 @@
-"@vtex/prettier-config"
+"@vtex/prettier-config"
```
danzanzini, line 1:
```suggestion
"@vtex/prettier-config"

```

danzanzini (approved):
Nice improvement. Code looks good to me 🚀

PR #607, opened 2022-10-13
[USAAPPTEAM-726] Fix ShowMoreFilterButton href attr

polishq:

A US client has noticed a discrepancy between the canonical URL of their custom search pages, i.e.:

`https://www.budgetgolf.com/category/all-golf-footwear?page=2`

and the URL used as the `href` attribute on the "Show More" button, i.e.:

`https://www.budgetgolf.com/category/all-golf-footwear?page=2&map=department`

This mismatch may be causing Google indexing issues.

This PR prevents the `map` query string from being added to the href attribute if `map` is equal to "department". 

#### How to test it?

Linked here: https://arthur--budgetgolf.myvtex.com/category/all-golf-footwear

### DIFF
--- a/react/components/loaders/FetchMoreButton.js
+++ b/react/components/loaders/FetchMoreButton.js
@@ -27,7 +27,7 @@ const useShowButton = (to, products, loading, recordsFiltered) => {
 }
 
 function shouldNotIncludeMap(map) {
-  if (!map || map === 'b' || map === 'brand') {
+  if (!map || map === 'b' || map === 'brand' || map === 'department') {
     return true
   }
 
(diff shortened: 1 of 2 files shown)


react/components/loaders/FetchMoreButton.js
```diff
@@ -27,7 +27,7 @@ const useShowButton = (to, products, loading, recordsFiltered) => {
 }
 
 function shouldNotIncludeMap(map) {
-  if (!map || map === 'b' || map === 'brand') {
+  if (!map || map === 'b' || map === 'brand' || map === 'department') {
```
thalytafabrine, line 30: I think you should include `map === 'c'` as well, since in most cases the map for department filter is `c`. `category-1` is also a possible option

polishq: Added!

PR #608, opened 2022-10-21
Remove map from FetchMore / FetchPrevious buttons if not in current URL

polishq:

For clients who use custom search pages in their stores, the `?map` query parameter is not usually part of their URLs, but it shows up in the `href` attribute of the Fetch More / Fetch Previous buttons, and this mismatch can cause issues with the Google crawlability of the site.

For example, this page: https://budgetgolf.myvtex.com/closeout/apparel
has a fetch more button with an `href` of https://budgetgolf.myvtex.com/closeout/apparel?page=2&map=department,promotion

This PR solves the issue by removing the `map` parameter from the button href if it does not appear in the current URL.  

#### How to test it?

New version linked here: https://arthur--budgetgolf.myvtex.com/closeout/apparel


thalytafabrine (approved):
lgtm

PR #610, opened 2022-11-07
Feature/add pricerange to the siteeditor

brendacdMoura:
When you set a price range via the site editor, the priceRange filter still appears among the filters' options.

(https://vtex-dev.atlassian.net/browse/IS-4144?atlOrigin=eyJpIjoiNGYzZmU1NmViMGVkNDBmYzkxYTI5ZjIyNjQ1NGZhYWMiLCJwIjoiaiJ9)

#### How to test it?

[Workspace](https://hiago--budgetgolf.myvtex.com/blank-test)

### DIFF
--- a/react/components/SearchQuery.js
+++ b/react/components/SearchQuery.js
@@ -141,7 +141,7 @@ const useCorrectSearchStateVariables = (
   return result
 }
 
-const useQueries = (variables, facetsArgs) => {
+const useQueries = (variables, facetsArgs, price) => {
   const { getSettings, query: runtimeQuery } = useRuntime()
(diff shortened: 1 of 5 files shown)


react/components/SearchQuery.js
```diff
@@ -164,6 +164,16 @@ const useQueries = (variables, facetsArgs) => {
     },
   })
 
+  const initialMap = () => {
```
hiagolcm, line 167: I would suggest to change the name. `initialMap` looks like an object instead of a function.

hiagolcm (approved):

PR #622, opened 2023-04-05
Add shipping options filter from session

thalytafabrine:
This is almost the same as the PR https://github.com/vtex-apps/search-result/pull/617 that [was yanked](https://github.com/vtex-apps/search-result/pull/621) due to an error when `selectedFacets` value was `undefined`. It fixes the error just by using an optional chaining operator

#### How to test it?

[Workspace](https://danieltest--auchan.myvtex.com/oferta-detergent-persil)


chrsmutti (approved):
LGTM

PR #641, opened 2023-09-29
Yank "enable ssr"

chrsmutti:

Some accounts have reported that after v3.126.1 their search-result
layout is breaking, we tracked to this commit and it's safe to remove
at the moment as it was just a step for another bigger commit that
was not merged still, we'll work on reverting this in the future.

This reverts commit 141ec3cd5fe2b7d93a4217a78d63b1a299d3dcea.

#### How to test it?

[Workspace](https://supp--lukshop.myvtex.com/outlet)

image


LucasTornai (approved):
LGTM

PR #643, opened 2023-10-20
Fix: Clean sponsored variable while the sponsored request is beign made

fltiago:

When a sponsored request takes too long and there is already a page change, the older sponsored product is shown. Which means sponsored products are being shown in weird contexts.

Example:
image

### DIFF
--- a/react/components/SearchQuery.js
+++ b/react/components/SearchQuery.js
@@ -188,13 +188,21 @@ const useQueries = (
     },
   })
 
-  const { data: { sponsoredProducts } = [] } = useQuery(
+  const { data: { sponsoredProducts } = [], loading: sponsoredProductLoading } = useQuery(
     sponsoredProductsQuery,
     {
       variables,
       skip: sponsoredProductsBehavior === 'skip',
     }
   )
+
+  const sponsoredProductReturn = () => {
+    if(sponsoredProductLoading) {
+      return []
+    }
+
+    return sponsoredProducts
+  }
 
   const {
     refetch: searchRefetch,
@@ -285,7 +293,7 @@ const useQueries = (
         sampling: facets?.sampling,
       },
       searchMetadata,
-      sponsoredProducts,
+      sponsoredProducts: sponsoredProductReturn(),
     },
     productSearchResult,
     refetch,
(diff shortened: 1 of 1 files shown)


react/components/SearchQuery.js
```diff
@@ -188,14 +188,22 @@ const useQueries = (
...
-  const { data: { sponsoredProducts } = [] } = useQuery(
+  const { data: { sponsoredProducts } = [], loading: sponsoredProductLoading } = useQuery(
     sponsoredProductsQuery,
     {
       variables,
       skip: sponsoredProductsBehavior === 'skip',
     }
   )
 
+  const sponsoredProductReturn = () => {
```
hcaula, line 199: I don't really like this function. Is not very common to do this in JavaScript, specially with it not receiving any params to justify its existence.

Also, you should also check for errors, not only the loading state.

I'd rather you'd just:

```javascript
const showSponsoredProducts = !sponsoredProductsLoading && !sponsoredProductsError
const sponsoredProductsResult = showSponsoredProducts ? sponsoredProducts : []
```

If you feel this is a level of abstraction below the current `useQueries` function (which I do), you can isolate this logic of requesting sponsored products into a different function.

react/components/SearchQuery.js
```diff
@@ -188,14 +188,22 @@ const useQueries = (
     },
   })
 
-  const { data: { sponsoredProducts } = [] } = useQuery(
+  const { data: { sponsoredProducts } = [], loading: sponsoredProductLoading } = useQuery(
```
hcaula, line 191: By the way, in order to get the error state as I commented below, you can do:

```suggestion
  const { data: { sponsoredProducts } = [], loading: sponsoredProductsLoading, error: sponsoredProductsError } = useQuery(
```

After that, the error will be available in the `sponsoredProductsError` variable.

hcaula (changes requested):
Please, also make sure this does not break the `exitocol` account.

Switch into `exitocol` and link this app. Search for "televisores" and open the network tab to check if the sponsored products are still being requested.

react/components/SearchQuery.js
fltiago: Ok, I didn't like the `! && !`, neither the ternary `if`, but as you said, seems to be a prevalent thing in JS.

hcaula (approved):

PR #645, opened 2023-11-17
feat: add gallery elementToScrollTo behavior

Susanlotuss:

The change made involves adding the 'scrollToElement' behavior to the Gallery component, removing the prop 'lazyItemsRemaining' in both the Gallery component and GalleryLayout. Additionally, 'lazyRender' has been removed from the GalleryLayoutRow component. This modification addresses a request wherein, when a user is in the PLP and clicks on a product to move to the PDP, then returns to the PLP, the user remains at the initial position. This ensures that the user does not need to scroll to find the position they were in before navigating to the PDP.

#### How to test it?

This behavior can be tested at: https://renwtesting--renwil.myvtex.com 

elementToScrollTo

#### Related to / Depends on

It is important to mention that the removal of lazyRender and lazyItemsRemaining from the initially mentioned components is due to the need to position the scroll in the PLP at the correct location where the user was before moving to the PDP. An ID is assigned to each item (GalleryLayoutRow component) for this purpose. This ID is captured during the click event to the PDP and retrieved upon returning to the PLP to locate the corresponding ID and scroll to that position. Therefore, it is necessary for these IDs to have already loaded; otherwise, the ID won't be found until manual scrolling is performed. Hence, the removal of lazy rendering.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,6 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ### changed
 - Behavior `scrollToElement` added to Gallery component for PLP.
-- `lazyItemsRemaining` removed in Gallery component and GalleryLayout. `lazyRender` removed in GalleryLayoutRow component.
 
 ### Changed
 - Bump `vtex.store-resources` version.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [3.127.1] - 2023-10-24
 
+### changed
+- Behavior `scrollToElement` added to Gallery component for PLP.
+- `lazyItemsRemaining` removed in Gallery component and GalleryLayout. `lazyRender` removed in GalleryLayoutRow component.
+
 ### Changed
 - Bump `vtex.store-resources` version.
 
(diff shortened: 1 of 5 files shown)


thalytafabrine (commented):
lgtm, but I have some questions before approving:
1. have you tested what can happen in stores that currently use the search rendering optimization option?  
2. we also have the lady render in the GalleryLegacy component, shouldn't we remove it from there as well?
3. will the "search rendering optimization" option be removed from the admin as we will no longer have this feature?

Susanlotuss:
> lgtm, but I have some questions before approving:
> 
> 1. have you tested what can happen in stores that currently use the search rendering optimization option?
> 2. we also have the lady render in the GalleryLegacy component, shouldn't we remove it from there as well?
> 3. will the "search rendering optimization" option be removed from the admin as we will no longer have this feature?

Hello! I've been analyzing the questions you posed, and I observed that it's possible to avoid the change related to the removal of lazy rendering. Instead, each store will have the option to disable the 'Enable lazy rendering of search results and facets' from the admin panel under the path Store Settings > Storefront > Store > Advanced. This way, the automatic scrolling on the PLP will work. Otherwise, if this option is enabled, it will function as it currently does, without the automatic scrolling on the PLP taking any effect. In light of this, I have already implemented the necessary changes in the last commit.

This behavior can be tested at: https://renwtesting--renwil.myvtex.com as the option 'Enable lazy rendering of search results and facets' has already been disabled in the admin for this store.

thalytafabrine (approved):
thanks for the great contribution @Susanlotuss 🤗

thalytafabrine:
hey @Susanlotuss  we had to revert this changes since it generated a bug on our stores. 
we will try to think about the use case you mentioned here and implement a different solution later :)

Susanlotuss:
> hey @Susanlotuss we had to revert this changes since it generated a bug on our stores. we will try to think about the use case you mentioned here and implement a different solution later :)

Hello, I understand, may I ask what the bug is? Thank you for letting me know!

PR #646, opened 2023-11-22
Feature: use settings to determine whether or not to fetch ads

hcaula:

In order to stop requiring changes to the store theme for fetching ads, we now use the setting `fetchSponsoredProductsOnSearch`.

#### How to test it?

Use the workspace below and log `window.__RUNTIME__.settings['vtex.store']` in your console. You should find the setting `fetchSponsoredProductsOnSearch`.

[Workspace](https://oneclickads--sjdigital.myvtex.com/cotonete?_q=cotonete&map=ft)

image

#### Related to / Depends on

Depends on:
- [ ] https://github.com/vtex-apps/admin-pages/pull/456
- [ ] https://github.com/vtex-apps/store/pull/581

### DIFF
--- a/react/components/SearchQuery.js
+++ b/react/components/SearchQuery.js
@@ -169,6 +169,12 @@ const useCorrectSearchStateVariables = (
   return result
 }
 
+const skipSponsoredProducts = ({ sponsoredProductsBehavior, settings }) => {
+  const fetchSponsoredProductsConfig = settings?.fetchSponsoredProductsOnSearch
+
+  return !fetchSponsoredProductsConfig && sponsoredProductsBehavior === 'skip'
+}
+
 const sponsoredProductsResult = (
   sponsoredProductsLoading,
   sponsoredProductsError,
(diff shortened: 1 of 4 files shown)


react/components/SearchQuery.js
```diff
@@ -169,6 +169,12 @@ const useCorrectSearchStateVariables = (
   return result
 }
 
+const skipSponsoredProducts = ({ sponsoredProductsBehavior, settings }) => {
+  const fetchSponsoredProductsConfig = settings?.fetchSponsoredProductsOnSearch
+
+  return !fetchSponsoredProductsConfig && sponsoredProductsBehavior === 'skip'
```
hcaula, line 175: It's still necessary to check for the `sponsoredProductsBehavior` value, since the account `exitocol` still won't abide to these changes. Once they do, we can remove this.

| Account state | Value of `fetchSponsoredProductsOnSearch` | Set the `sponsoredProductsBehavior`? | Boolean outcome | Will fetch ads? |
|--|--|--|--|--|
| Normal account | `false` | `false` |  `false` | No |
| VTEX Ads account | `true` | `false` | `true` | Yes |
| exitocol | `false` | `true` | `true` | Yes |

thalytafabrine (approved):

react/components/SearchQuery.js
hcaula: @thalytafabrine Yes, of course!

PR #654, opened 2024-02-05
fix: decreasing timer to be able to scroll down the category

luizfelipebraga:

Bugfix: Scroll down after clicking a product and go back to category.

The issue manifests when attempting to scroll through a category page immediately after clicking on a product within that category and then navigating back to the category page. We've received reports from several stores highlighting this bug and its adverse effects on the overall user experience within the store. 

[Video showing the bug](https://www.loom.com/share/d174009905834a829d13fd051ac52559?sid=98d2c9a2-2023-4b85-a1f8-dacdb371ea6f)

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -1,7 +1,7 @@
 {
   "name": "search-result",
-  "vendor": "aleatory",
-  "version": "3.129.6",
+  "vendor": "vtex",
+  "version": "3.129.5",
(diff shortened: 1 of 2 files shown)


thalytafabrine:
closing since it was solved [here](https://github.com/vtex-apps/search-result/pull/659)

PR #658, opened 2024-03-04
Get sponsored products in productSearch

thalytafabrine:
#### How to test it?

[workspace](https://thalyta--lebiscuit.myvtex.com/fritadeira?_q=fritadeira&map=ft)


hcaula (approved):
I liked the `sync` and `async` options!

thalytafabrine:
> I liked the `sync` and `async` options!

@hcaula great! I was in doubt about these names hahah not very creative and I was thinking if this could get confused.. but since you liked I think we can follow with them :)

PR #661, opened 2024-04-19
Refactor: remove explicit sponsored product call

hcaula:

This PR does two things:

- Removes the ability to call sponsored products separately from the product search request;
- Always sends the param `showSponsored` as `true` to the product search request.

Now, we check in the intelligent-search-app if the account supports sponsored products.

#### How to test it?

- [farmadelivery](https://caula--farmadelivery.myvtex.com/dipirona?_q=dipirona&map=ft) - This account used to call ads separately from the product search - now, it's only one request.

#### Related to / Depends on

- [x] https://github.com/vtex/intelligent-search-api/pull/100


react/components/SearchQuery.js
```diff
@@ -199,31 +177,11 @@ const useQueries = (
   const productSearchResult = useQuery(productSearchQuery, {
     variables: {
       ...variables,
-      showSponsored:
-        settings?.fetchSponsoredProductsOnSearch &&
-        sponsoredProductsBehavior === 'sync',
+      showSponsored: true,
```
thalytafabrine, line 180: if this flag will always be `true` and we check the store in the resolver, can't we just remove the flag here?

hcaula: @thalytafabrine Not yet, because not all placements support ads yet, like shelves and autocomplete. But both of them call the `productSearch` query.

thalytafabrine (approved):

PR #664, opened 2024-04-29
Add Hungarian translations

GeorgeLimaDev:
#### What problem is this solving?

It relates to the task [LOC-14426](https://vtex-dev.atlassian.net/browse/LOC-14426). Adds translations for Hungarian (`HUN` locale).

[LOC-14426]: https://vtex-dev.atlassian.net/browse/LOC-14426?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #665, opened 2024-05-06
Fix Hungarian translations

GeorgeLimaDev:

It fixes this previous PR #664 that had reserved words that should not have been translated.


chrsmutti (approved):
LGTM

PR #671, opened 2024-10-01
fix: duplicated structured data in product search context

iago1501:
### What problem is this solving?

This commit has the intention to fix the duplicated array of itemList in structured data , that occurs on PLP pages. Google rich result is complaining about this duplicated node element causing some crawling issues

We found out that there are two occurrences of ProductListStructuredData in the same PLP.
The [SearchWrapper](https://github.com/vtex-apps/store/blob/96b16702d2ff78291b3f53224504dcbd9803959d/react/SearchWrapper.tsx#L450) element from `vtex-apps/store` is responsible for the first. `Gallery` and `vtex.search-result` is responsible for [the other](https://github.com/vtex-apps/search-result/blob/600e4a9c0f4761f940867ff5846ae21121da6d1d/react/Gallery.tsx#L32) 

What happens is that when in search context we add the structured data, and, in search context, we also use gallery, which is also adding the structured data.

The main purpose is to check if the block is under the context of store.search to check if it's going to add the structured data element or not

This PR simply adds this validation to understand in which context the gallery is being used to avoid duplicated node content

### How to test it?

[Before](https://storetheme.vtex.com/apparel---accessories/)

**Steps to reproduce:**

* Open the link above
* Open DevTools
* Go to Elements tab
* Search (cmd+f) for the following string: `{"@context":"https://schema.org","@type":"ItemList","itemListElement"`

**Expected Behavior:**

* You should find two occurrences of this string
* The content of the Structured JSON that starts with the searched string must be the same (duplicates)
image

[After](iespinoza--storecomponents.myvtex.com/apparel---accessories/)

**Steps to reproduce:**

* Open the link above
* Open DevTools
* Go to Elements tab
* Search (cmd+f) for the following string: `{"@context":"https://schema.org","@type":"ItemList","itemListElement"`

**Expected Behavior:**

* You should find only one occurrence of this string. 
  * The first occurrence in the HTML, from the `SearchWrapper` component
  
image

We adopted this approach instead of removing it from store.search the gallery component just to be safe, avoiding any undesired behavior since we don't know exactly the side effects that this might cause

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+- Duplicated structured data in product search context
+
 ## [3.133.0] - 2024-08-01
 
 ### Added
--- a/react/Gallery.tsx
+++ b/react/Gallery.tsx
@@ -1,5 +1,6 @@
 import React, { Fragment } from 'react'
 import { ProductList as ProductListStructuredData } from 'vtex.structured-data'
+import { useRuntime } from 'vtex.render-runtime'
 
 import GalleryLayout from './GalleryLayout'
 import type { GalleryLayoutProps, Slots } from './GalleryLayout'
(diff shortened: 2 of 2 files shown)


vmourac-vtex (approved):
LGTM!

PR #675, opened 2024-10-03
feat: add facets default to the shipping filter

ArthurFerrao:
Show delivery facets even when CEP is not defined.
[SAX-334](https://vtex-dev.atlassian.net/browse/SAX-334)

#### How to test it?

Search for a product (example query: "air"), without a defined CEP, and see the Shipping filter in the sidebar.

[Workspace](https://arthurdp--fulfillmentqa.myvtex.com/air?_q=air&fuzzy=0&initialMap=ft&initialQuery=air&map=ft&operator=and)

image


react/components/FacetCheckboxListItem.js
```diff
@@ -1,5 +1,5 @@
 import classNames from 'classnames'
-import React from 'react'
+import React, { useMemo } from 'react'
```
GabEluan, line 2: Is this new import being used?

react/utils/getFilters.js
```diff
@@ -74,4 +103,26 @@ const getFilters = ({
   ].filter(Boolean)
 }
 
+const getFormattedDeliveries = (deliveries, showShippingFacet) => {
+  if (!showShippingFacet) {
+    return deliveries.filter(d => d.name !== SHIPPING_KEY)
+  }
```
GabEluan, line 109: Just checking, this is to support facets other than `shipping` in the future for Delivery Promises, correct?

GabEluan (approved):
Beautiful. Left two minor comments.

luarakerlen (approved):

react/utils/getFilters.js
ArthurFerrao: Correct, to ensure I'm not removing other facets that can be in this array

react/components/FacetCheckboxListItem.js
ArthurFerrao: Yes, this is. I don't know how this file worked before, because the useMemo was in use(line 29) but the import didn't exist. It only broke when I used the filter in mobile mode.

PR #690, opened 2024-11-29
Add missing scrollToTop prop to accordion filter

llfalcao:

The `scrollToTop` prop isn't passed down to the `PriceRange` component when the collapsible navigation type is applied to the filter navigator, causing the window to scroll to the top when query params are updated.

#### How to test it?

Using any component that updates the query params with `setQuery`, from [render-runtime](https://github.com/vtex-apps/render-runtime).

[Workspace (before)](https://mobilefilter--compracerta.myvtex.com/pecas-e-acessorios)
[Workspace (after)](https://mobilefilterfix--compracerta.myvtex.com/pecas-e-acessorios)

Before

[Before](https://github.com/user-attachments/assets/9daf72d5-d323-4892-821a-cf7f443c3fe3)

After

[After](https://github.com/user-attachments/assets/1a7113da-6265-4982-9f8a-b1e359a6fc5e)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 - Updated English, Spanish and Portuguese translations.
 - Added new strings to remaining languages.
+- `scrollToTop` prop not being passed down to the `PriceRange` component.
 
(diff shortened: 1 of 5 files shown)


llfalcao:
I can't check any of these, but it's a patch and there's nothing new to document.

image

image

PR #758, opened 2026-06-01
fix: disable pickup-in-point filter when facet quantity is 0

hiagolcm:
## Problem

The `pickup-in-point` radio option in the filter navigator was appearing as **active/enabled** even when the facets API returned `quantity=0` for it — meaning no results were available for that shipping option.

This happened because the disabled logic for `pickup-in-point` only checked whether there were pickup points in the `delivery-promise-components` context (`pickups.length === 0`), completely ignoring `facet.quantity`. Since `compatibilityLayer.js` always injects `pickup-in-point` as a default with `quantity: 0` when the API doesn't return it, the filter appeared clickable even when it had no results.

All other filters (`delivery`, `pickup-nearby`, etc.) correctly use `facet.quantity === 0` to determine their disabled state.

## Fix

**`react/components/RadioFilters.js`**

```diff
- disabled: isPickupInPointShippingValue(facet.value)
-   ? pickups.length === 0
-   : facet.quantity === 0,
+ disabled: isPickupInPointShippingValue(facet.value)
+   ? pickups.length === 0 || facet.quantity === 0
+   : facet.quantity === 0,
```

`pickup-in-point` is now disabled when **either**:
- The facets API returned `quantity=0` (no products available for that filter), **or**
- There are no pickup points in the delivery-promise context

## Tests

Added `react/__tests__/RadioFilters.test.js` with 19 tests covering:

- `pickup-in-point` disabled when `quantity=0` even with pickups in context
- `pickup-in-point` disabled when no pickups in context even with `quantity > 0`
- `pickup-in-point` enabled only when both `quantity > 0` AND pickups exist
- Same rules apply to `pickup-in-point-{id}` (URL-selected variant)
- `delivery` and `pickup-nearby` disabled only by `quantity=0` (unaffected by pickups)
- Mixed facet scenarios
- Interaction: `onChange` called for enabled facets, not for already-selected ones

## Test plan

- [ ] Verify the filter appears **disabled** on a PLP where pickup-in-point is not returned by the facets API
- [ ] Verify the filter appears **enabled** on a PLP where pickup-in-point is returned with `quantity > 0` and there are pickups in context
- [ ] Verify `delivery` and `pickup-nearby` behavior is unchanged

Made with [Cursor](https://cursor.com)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+- `RadioFilters`: `pickup-in-point` option is now disabled when the facets API returns `quantity: 0`, even if pickup points exist in the delivery-promise context. Previously the disabled check only looked at `pickups.length`, causing the filter to appear active when it had no actual results (e.g. when injected as a default placeholder by `compatibilityLayer`). Now matches the same contract as all other filters: disabled when either `quantity === 0` OR (for pickup-in-point specifically) `pickups.length === 0`.
+
 ## [3.147.1] - 2026-06-02
 
 ### Added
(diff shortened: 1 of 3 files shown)


iago1501 (approved):
Nice catch!

PR #760, opened 2026-06-11
Feat/zipcode soft refresh query allowlist

hiagolcm:
#### How to test it?

[Workspace](Link goes here!)


react/__tests__/useFetchMore.test.js
```diff
@@ -0,0 +1,91 @@
...
+
+  expect(result.current.nextPage).toBe(6)
+
+  // The shopper clicks "load more": the URL page advances in lockstep — no reset.
+  setRuntimePage(6)
+  act(() => {
+    rerender({ ...baseProps, page: 5 })
+  })
+
+  expect(result.current.nextPage).toBe(6)
```
chrsmutti, line 79: 🟡 

Shouldn't this be 7? 🤔

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## [Unreleased]
 
+### Fixed
+
+- `useFetchMore` (gallery "load more"): the in-memory pagination reducer now resets to the first page when the `page` query param is **externally** brought back to page 1 (e.g. `vtex.delivery-promise-components` clearing it through render-runtime after a zipcode / pickup-point change). Previously, because the reducer is seeded once at mount and only resets on `query`/`map`/`orderBy`/`priceRange` changes, a location change left it on the stale page: a shopper on page 5 would see page-1 results but the next "load more" fetched page 6 (and the URL showed `page=6`). The reset fires only on an external transition to the first page while the reducer is past it, so the shopper's own "load more"/"fetch previous" navigation and initial mount are unaffected.
```
chrsmutti: 🟡 

I'd prefer to make this cleaner, single sentences. And move the "Why?" to the PR description.

chrsmutti (approved):
I'm not very familiar with the lazyItems stuff, but it looks right! Just have a nitpick on the Changelog.

react/__tests__/useFetchMore.test.js
hiagolcm: It should be 6. This is a safeguard test.

It would only be 7 if handleFetchMoreNext were called, but this test only changes the URL, which shouldn't cause any state change. The previous test covers the reset that happens when the URL goes to page undefined; this one guarantees that changing the URL to any other number does nothing. It's basically testing the "false" path of if (page === undefined) { reset }.

PR #767, opened 2026-06-25
TIS-710 Fix data-af-product-position on paginated PLP loads

thiagopereira-vtex:
## Summary

- Fix `data-af-product-position` so gallery items report their true rank in the search result set, not just their index within the currently rendered DOM
- Introduce `ProductsPositionOffsetContext` fed by `useFetchMore`'s live `from` offset, covering direct page loads (`?page=2`) and fetch-previous scenarios
- Apply the corrected position in both `GalleryRow` and `GalleryLayoutRow`, including pixel `productClick` events via the `position` prop

Closes [TIS-710](https://vtex-dev.atlassian.net/browse/TIS-710)

## Test plan

- [x] `make check` (lint + 159 unit tests)
- [x] Search on a store with show-more pagination (e.g. Intelbras, 18 items/page)
- [x] Verify positions 1–18 on page 1
- [x] Click "Load more" → positions 1–36
- [x] Reload at `?page=2` → first product shows `data-af-product-position="19"`
- [x] From page 2, click "Fetch previous" → positions 1–36 with correct sequence
- [ ] Repeat on flexible layout (`search-result-layout` + gallery layout options)

[TIS-710]: https://vtex-dev.atlassian.net/browse/TIS-710?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #772, opened 2026-07-06
TIS-710 Fix data-af-product-position on direct ?page=N loads

thiagopereira-vtex:
## Summary

- Follow-up to #767 / `3.148.1`: the initial fix introduced `ProductsPositionOffsetContext`, but the offset still stayed at `0` on direct `?page=2` reloads because `useFetchMore` seeded `from` from the `page` prop instead of the URL, and the reducer only initializes once
- Seed `useFetchMore.from` from `currentPage` (`?page=` → prop → default)
- Add `getProductsPositionOffset` to resolve the gallery offset from `searchQuery.variables.from` / URL when only a single page is in the DOM, while keeping offset `0` for load-more and fetch-previous flows

Closes [TIS-710](https://vtex-dev.atlassian.net/browse/TIS-710)

## Test plan

- [x] `make check` (lint + 177 unit tests)
- [ ] Reload PLP at `?page=2` (18 items/page) → first product has `data-af-product-position="19"`
- [ ] Page 1 → load more → positions remain 1–36
- [ ] Page 2 → fetch previous → positions remain 1–36 starting from 1
- [ ] Verify on flexible layout (`search-result-layout`)

[TIS-710]: https://vtex-dev.atlassian.net/browse/TIS-710?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


hiagolcm:
Why does the from value derived from the page prop still equal 0 on direct ?page=2 reloads? You said that 'the reducer only initializes once,' but I don't quite understand the whole flow yet.

thiagopereira-vtex:
> Why does the from value derived from the page prop still equal 0 on direct ?page=2 reloads? You said that 'the reducer only initializes once,' but I don't quite understand the whole flow yet.

Good question — there are two separate pagination offsets on the PLP that weren’t aligned.

SearchQuery reads ?page= from the URL and sets searchQuery.variables.from correctly, so the GraphQL query fetches the right products on a direct ?page=2 reload. PR #767, however, wired the gallery position to pageState.from from useFetchMore — a different value, managed by an internal reducer for load more / fetch previous.

Before this PR, that reducer seeded from from the page prop, which in the flexible layout often doesn’t reach SearchResultContainer (undefined or 1), even when the URL already has ?page=2. At the same time, other fields like to were derived from the URL. Result: the query fetched products 19–36, but the position offset stayed at 0.

The reducer also only reads its initial state on mount. Once from is set to 0, it doesn’t update until the shopper interacts (load more or fetch previous). On a direct reload there’s no interaction, so from stays 0 for the whole session.

This PR fixes the reducer seed to use the URL page, and adds a fallback to searchQuery.variables.from / ?page= when the reducer hasn’t caught up yet and only one page is rendered.

hiagolcm (approved):

PR #775, opened 2026-07-16
feat: unify delivery-promise facet visibility

hiagolcm:
[Workspace](https://unifyvisibility--vinotecamx.myvtex.com/)

## Summary (written by human)
The delivery promise related facet will only be shown in two situations

1. Merchant deliberately set the `showDeliveryPromiseFacets` to true
2. or, there is at least one delivery-promise-component block (shopper-location-setter, pickup-point-selector or shipping-method-selector) rendered in the page

This PR also keep retrocompatibility with `showShippingFacet` and `showShippingMethodFacet`.

## Changes
- add `showDeliveryPromiseFacets` as the single gate for `shipping`, `dynamic-estimate`, and `delivery-options`
- preserve deprecated `showShippingMethodFacet` as a fallback when the new prop is unset
- infer default visibility from mounted delivery-promise blocks in the shared `uiRegistry`
- apply the same resolved gate to facet groups and selected-filter badges
- document the new prop, precedence, and deprecation behavior

## Tests
- added pure truth-table tests for prop precedence and block-presence fallback
- added selected-filter coverage for all three delivery-promise keys
- updated `getFilters` tests for the unified gate and `hiddenFacets` interaction
- full React suite: 193 tests passed
- all changed files pass ESLint; pre-commit lint and formatting hooks passed

## Assumptions
- removing the old `showShippingMethodFacet = false` default is required to distinguish an omitted prop from an explicit `false`
- selected-filter badges follow the same unified gate as their facet groups

## Dependencies
- https://github.com/vtex-apps/delivery-promise-components/pull/34

## Spec
- https://github.com/vtex-apps/is-io-specs/pull/14

Made with [Cursor](https://cursor.com)


chrsmutti (approved):
LGTM

julia-rabello (approved):

